PR #17, opened 2018-07-18
Moves your form.js and HTML to Next pages

craigspaeth:
Moves your static HTML into [Next.js](https://github.com/zeit/next.js/) React components under a /pages directory (Next.js convention). I'm hoping this looks similar enough to what you wrote that it's easy to grok because it allows us to easily leverage Next's build tooling vs. the whole separate Webpack thing you had set up.

...but if not, I can look into seeing if there's an easy way to use Next's version of wepback to do a more "traditional" static web build.


apps/home/index.js
```diff
@@ -1,7 +1,20 @@
...
+
+const nextHandler = async (req, res) => {
+  await prepare()
+  handle(req, res)
+}
 
-app.use('/', express.static(`${__dirname}/static`))
-app.get('/', (req, res) => res.send('Hello World!'))
+app.get('/', nextHandler)
+app.get('/static*', nextHandler)
```
craigspaeth: You don't really have to worry about this server-side part as I don't imagine you'd touch this stuff usually.

apps/home/pages/_document.js
```diff
@@ -0,0 +1,50 @@
...
+          />
+          <link
+            rel='icon'
+            type='image/png'
+            sizes='16x16'
+            href='https://goligorsky.neocities.org/Octave_LP_07082018/favicon-16x16.png'
+          />
+        </Head>
+        <body>
+          <Main />
```
craigspaeth: As you might have noticed, I've copy-pasted your `<head>` and `<script>` HTML into this [`_document`](https://github.com/zeit/next.js/#custom-document) page which is Next's convention for the "layout" or "root level react component".

The `<Main />` tag here tells Next that this is the place to inject other Next "pages" (the index.js react component below).

apps/home/pages/_document.js
```diff
@@ -0,0 +1,50 @@
...
+          <script src='https://ajax.googleapis.com/ajax/libs/jquery/1.11.3/jquery.min.js' />
+          <script src='https://static.airtable.com/js/embed/embed_snippet_v1.js' />
+          <script src='/static/scrolling.js' />
+          <NextScript />
```
craigspaeth: This is just another Next React component that injects Next's secret sauce javascript.

apps/home/pages/_document.js
```diff
@@ -0,0 +1,50 @@
...
+            type='image/png'
+            sizes='16x16'
+            href='https://goligorsky.neocities.org/Octave_LP_07082018/favicon-16x16.png'
+          />
+        </Head>
+        <body>
+          <Main />
+          <script src='https://ajax.googleapis.com/ajax/libs/jquery/1.11.3/jquery.min.js' />
+          <script src='https://static.airtable.com/js/embed/embed_snippet_v1.js' />
+          <script src='/static/scrolling.js' />
```
craigspaeth: As you can see here I prepended `src` and `href`s to have `/static/`—which is [just another Next convention](https://github.com/zeit/next.js/#static-file-serving-eg-images).

apps/home/pages/index.js
```diff
@@ -0,0 +1,153 @@
...
+              <iframe
+                className='airtable-embed airtable-dynamic-height'
+                src='https://airtable.com/embed/shr13Mne0It2ltmH0?backgroundColor=purple'
+                frameBorder='0'
+                width='100%'
+                height='1066'
+              />
+            </div>
+            <div>
+              <TextField
```
craigspaeth: Here's where I moved your form.js code. Because the rest of your HTML was simply copy-pasted into a React component, you can now just include the material react components without any of the other `render` boilerplate.

PR #51, opened 2018-07-30
added checkbox and related form control components to index.js

dGoligorsky:


apps/prelaunch/pages/index.js
```diff
@@ -29,9 +38,14 @@ export default class Body extends React.Component {
     })
   }
 
+  handleChange = name => event => {
```
craigspaeth: 👌 nice handling of form elements with state here. Nit-pick, but `handleChange` might be a little too generic sounding. Not blocking the PR for it though. thanks!

PR #68, opened 2018-08-08
Update consult schedule flow

craigspaeth:
Updates consult schedule flow to be in line with latest wireframes, including an editable "consent" page.

consult


dGoligorsky:
🤣 Nice checkbox options

PR #85, opened 2018-08-14
Submitted state for lead gen form so it doesn't look broken

craigspaeth:
I want to get a deploy out and I realized the form needs a submitted state to not look broken. So I did some very basic copy change on submit. Feel free to change it or lmk if you'd like something different.

cap


dGoligorsky:
I was thinking the button would take on a submitted state (disabled) so the button text would be "Thanks! We'll be in touch." and would be un-clickable. And the input fields would be the disabled state but keep the user-entered data in them.

If that's too much of a pain, can we at minimum do an opacity transition so the form fades out and the text fades in? Might make it less jarring from state to state.

PR #86, opened 2018-08-14
Less jarring submit states

craigspaeth:
cap


dGoligorsky:
I think this is cleaner! It'd be even better if the input fields were disabled too, so you could see what you submitted but after submit you can't edit the fields...

PR #112, opened 2018-08-24
Change EO buttons to orange

craigspaeth:
Uses orange as the secondary MUI color palette and updates the buttons to orange. Apologies for the noise in the diff—that's from the auto-formatting that our build runs.
image


lib/next-mui.js
```diff
@@ -31,9 +30,10 @@ const theme = createMuiTheme({
       dark: colors.secondaryDarkBlue
     },
     secondary: {
-      light: green[300],
-      main: green[500],
-      dark: green[700]
+      light: colors.secondaryLightOrange,
+      main: colors.secondaryLightOrange,
```
craigspaeth: I didn't see any uses of "secondary" in the project yet, so I assumed it would be okay that I replace the copy + pasta'd `green` here with our oranges.

apps/prelaunch/pages/eocl1.js
```diff
@@ -73,7 +73,7 @@ export default class Body extends React.Component {
           type='submit'
           fullWidth
           variant='contained'
-          color='primary'
+          color='secondary'
```
craigspaeth: This MUI attribute  is what controls the color on the individual buttons.

craigspaeth:
Mergin' to avoid conflict with EO scripts

PR #118, opened 2018-08-24
EO thank you page. Tracking script component is on there.

dGoligorsky:
What's the smart way that people mask email addresses these days from robots?


craigspaeth:
Good question about email masking. These days bots are rather sophisticated and sometimes it's a [feature not a bug](https://www.urbandictionary.com/define.php?term=It%27s%20not%20a%20bug%2C%20it%27s%20a%20feature) in the case of e.g. screen readers. So it's better to just not worry about it and rely on spam filters to do their things [read more](https://stackoverflow.com/questions/41318987/hide-email-address-from-bots-keep-mailto).

craigspaeth:
LGTM, thanks!

PR #123, opened 2018-08-27
added illos and coaching page ready to go

dGoligorsky:


craigspaeth:
LGTM! Although I don't understand the `illo` is that a way of avoiding name conflicts?

dGoligorsky:
Yeah, basically. I should probably be more strict about my naming, actually. Probably more fast + loose with this marketing test page, otherwise I'd be changing all the class names to something better.

craigspaeth:
👍 np, just curious. We're using some [cool tools](https://github.com/css-modules) to help make that namespacing easier in the rest of the project. I'm fine with fast + loose CSS in this prelaunch folder.

PR #125, opened 2018-08-27
EO configuration

craigspaeth:
Hooks up the EO marketing pages correctly to lead w/ mailing list generation. Also fixes a small issue with meta descriptions not wiring up correctly.


apps/api/models/lead.js
```diff
@@ -20,7 +20,7 @@ const fields = (method = 'read') =>
   email: String${method === 'create' ? '!' : ''}
   ${{ create: '', read: 'id: ID', update: 'id: ID!' }[method]}
   name: String${method === 'create' ? '!' : ''}
-  phone: String${method === 'create' ? '!' : ''}
+  phone: String
```
craigspaeth: Making phone # not required for now b/c these pages don't include them, but I think ideally we'd want to make sure this is required as many places in the product need it for e.g. consult calls/coaching/2fa/etc.

PR #128, opened 2018-08-27
sync assessments job, envenc update

lionsdwarf:


apps/jobs/lib/sync-assessments.js
```diff
@@ -0,0 +1,147 @@
...
+          airtableRecordId: record.metadata.id
+        }
+      })
+    }
+  }
+}
+
+const runSync = async function () {
+  const patients = await cliniko.find({ resource: `patients` })
+  patientsEmailToId = mapEmailToId(patients)
```
craigspaeth: It'd be better if this variable were assigned here and passed in as an arg to the following `syncATAssmntsToCliniko` calls. That avoids mutating a module-level variable which is good to avoid if possible.

At the moment it's not causing a lot of trouble but one could imagine us later trying to refactor this to run this job in parallel from a different module, e.g.

```javascript
import syncAssessments from 'lib/sync-assessments'

Promise.all([
  syncAssessments('registrationAssessment'), // 1
  syncAssessments('intakeScreenerAssessment'), // 2
  syncAssessments('PHQ-9'), // 3
  syncAssessments('etc') // 4
])
```

In that case job 2 could mutate `patientsEmailToId` while job 1 is looping through its records and bork job 1 mid-loop—otherwise called a race condition.

It'd be better to do:

```javascript
const patientsEmailToId = mapEmailToId(patients)
syncATAssmntsToCliniko('registrationAssessment', patientsEmailToId)
syncATAssmntsToCliniko('registrationAssessment', patientsEmailToId)
```

Not necessary here, but FWIW when passing things around like this gets clunky to read don't be afraid to use options objects or get liberal with using terse + contextually-relevant names, e.g.:

```javascript
const idsMap = mapEmailToId(patients)
syncATAssmntsToCliniko({ table: 'registrationAssessment', idsMap })
```

apps/jobs/lib/sync-assessments.js
```diff
@@ -0,0 +1,147 @@
...
+}
+
+const syncATAssmntsToCliniko = async function (atTable) {
+  const assmnts = await fetchUnsyncedATRecords(atTable)
+  const segregatedRecords = segregateMetadataFromContent(assmnts)
+  for (let record of segregatedRecords) {
+    if (patientsEmailToId[record.metadata.email]) {
+      let res = await cliniko.create({
+        resource: `treatment_notes`,
+        data: {
```
craigspaeth: This map of data passed to `create` is a good candidate to be pulled out to its own testable pure function e.g. 

```javascript
await cliniko.create({
  resource: 'treatment_notes',
  data: assessmentNote({ record, patientId: patientsEmailToId[record.metadata.email] })
})
```

Just FYI—not blocking the PR for it.

craigspaeth:
Awesome work! Mostly LGTM! Thanks 🙌

My comments are generally around encouraging a more functional or [immutable](https://medium.com/javascript-scene/the-dao-of-immutability-9f91a70c88cd) approach, but not blockers for this PR. Once the build is 💚 I'm happy to merge this (make sure to run `npm test` on your end). 

FWIW I find code more predictable and readable when you can avoid mutations and things like `delete`, `for` loops, and `let`. I encourage leveraging more `const` and passing data around functions while using [Lodash](https://lodash.com/docs/4.17.10) to shape data in the ways that you need. Happy to pair together on refactoring something to a more functional approach—if that sounds appealing lmk!

apps/jobs/test/sync-assessments.test.js
```diff
@@ -0,0 +1,109 @@
...
+    metadata: {
+      ID: `xxxx`,
+      id: `yyyy`,
+      name: `bob`,
+      email: `bob@corn.com`
+    }
+  })
+})
+
+test(`parseAirtableIdFromRes receives a response object and parses out the airtable id that was stamped on the cliniko note record`, () => {
```
craigspaeth: 👌 solid test coverage, thanks!

apps/jobs/lib/sync-assessments.js
lionsdwarf: Good point. Will do.

apps/jobs/lib/sync-assessments.js
lionsdwarf: Copy that. Will relocate variable assignation.

PR #130, opened 2018-08-28
Fetch & map therapist + coach profiles

craigspaeth:
Beginning of this page:

image

Provides data necessary for that view by allowing querying of the appointmentType > practitioners relationship in GraphQL and mapping that data into something view-friendly in a client-side "model"

image


README.md
```diff
@@ -40,7 +40,30 @@ Jest for unit and integration testing, Nightmare for end-to-end tests. Prettier-
...
+```
+practitioners(filter: {
+  forLeadId: "Karen",
+  byAppointmentType: THERAPY
+}) {
+  name
+}
+```
+
+When unsure how to design our schemas we'll look to popular GraphQL services like [GraphCMS](https://graphcms.com/) and [Primsa](https://www.prisma.io/) for inspiration. Given this "object oriented" structure we'll follow-that "Active Record-like" pattern here and encapsulate Apollo Server types and resolvers alongside business and persistence logic as modules under a /models folder. Because we can quickly get clients stuck using fields we expose, we'll strive for minimalistic CRUD-like coverage vs. exposing every GraphQL field out of the box. When evolving the API, aim to treat schemas as append-only, mark old fields deprecated, and only when necessary carefuly delete fields by migrating clients off and monitoring their usage. Authorization will use JWTs passed in the Authorization header to do role base auth logic which will simply throw auth errors in resolvers. Authentication will be handled by a service like Auth0.
```
craigspaeth: Small update to the docs to try and capture how I've thought about structuring GraphQL schemas/nested queries in this case. There's a lot of freedom/power in GraphQL that's still being explored and written about, so hoping to reflect some pragmatic guidelines here.

craigspaeth:
Going to merge, feel free to leave any comments if you'd like.

README.md
lionsdwarf: Copy that--I will dig into these links. Thanks for highlighting!

PR #131, opened 2018-08-28
Consolidate on CSS modules

craigspaeth:
This consolidates styling patterns on [CSS modules](https://github.com/css-modules)—dropping [styled jsx](https://github.com/zeit/styled-jsx) and [material ui](https://material-ui.com/) (sorry David, Matt & I were fighting it too much—I believe this is best for the long run), and in doing so involves a handful of refactors:

- Moves the MUI stuff to be only inside the ./prelaunch app, where it can continue to be used and eventually dropped with the whole prelaunch app come Oct. launch
- Enables absolute `@imports` on the lib folder like we do JS modules
- Moves some minimally-shared CSS vars like colors and typography to the ./lib folder to be shared across apps
- Drops the styled JSX and refactors the root-level button to use CSS modules
- Configures Jest to support the imports
- Adds CSS to the auto-formatting/linting pipeline with Prettier

cc @dGoligorsky @lionsdwarf


postcss.config.js
```diff
@@ -1,6 +1,9 @@
 module.exports = {
   plugins: [
-    require('postcss-easy-import')({ prefix: '_' }), // keep this first
+    require('postcss-easy-import')({
+      prefix: '_',
+      path: [`${__dirname}/lib`]
```
mkandler: Does this break all the imports of the `responsive.css` file in the marketing components? Should work to move that file in `/lib` and adjust the imports (I can do that after merging if you'd like)

apps/marketing/components/Layout/layout.css
```diff
@@ -1,12 +1,24 @@
 @import '../../styles/variables.css';
```
mkandler: Example of a place I think may be broken also... (see other comment)

dGoligorsky:
Ah crumbs. When back from vacation, I'll be curious to hear why MUI didn't work out and more about your alternative.

postcss.config.js
craigspaeth: I don't believe it breaks anything because it's just adding ./lib to the `@import` paths so we can `@import` into ./lib absolutely like [we do in JS](https://github.com/findOctave/cortex/blob/master/next.config.js#L12) e.g. `@import 'styles/typography.css'`—reason being I'm treating ./lib as this blessed directory that acts like a local node_modules.

Relative imports seem to be working fine:

cap

mkandler: Oh awesome, didn't realize how that worked, thanks for clarifying 😄

PR #133, opened 2018-08-28
UI for assessment zero state

craigspaeth:
Adds some UI boilerplate for the assessment zero state view, including adding [postcss-apply](https://github.com/pascalduez/postcss-apply) to get "native" CSS mixins for typography.

cap


lib/styles/index.css
```diff
@@ -1,15 +1,33 @@
...
   --pink-color: #f07694;
   --white-color: #fff;
-  --mostly-black-color: #232323;
+  --yellow-color: #eb991b;
 
   /* Typography */
-  --font-sans-serif: 'Apercu', 'Roboto', 'Work Sans', Helvetica, Arial,
-    sans-serif;
-  --font-serif: 'Esteban', 'Georgia', sans-serif;
+  --font-apercu: 'Apercu', 'Roboto', 'Work Sans', Helvetica, Arial, sans-serif;
```
craigspaeth: @mkandler I hope you don't mind that I renamed `serif` and `sans` to the typeface name. The reason being is in my experience there's a good chance that another sans/serif type family enters the mix at some point b/c of some edge case device/email-client/targeted-platform restrictions or simply the design team needing more typographical range. So I'm preferring to be more explicit here.

PR #137, opened 2018-08-29
cliniko records filtered to exclude records where appt ended within 4…

lionsdwarf:
…8 hrs and where patient did not arrive

@craigspaeth FYI--I did not run end-to-end test of this job, but instead just tested/confirmed data filtration


craigspaeth:
Going to merge and PR what I mean by that test.

PR #138, opened 2018-08-29
Signup stage notifications

lionsdwarf:
@craigspaeth -- I have not yet added logging, as discussed. Can add via separate ticket since it will apply to all Node jobs, not just this one. Let's discuss priority.


apps/jobs/lib/signupStage-notifications.js
```diff
@@ -0,0 +1,159 @@
+/**
```
craigspaeth: This capitalization on this filename is a bit odd. I'd prefer being consistently kebab-case `signup-stage-notifications.js`.

apps/jobs/lib/signupStage-notifications.js
```diff
@@ -0,0 +1,159 @@
...
+
+const LEADS_TABLE = `leads`
+
+const schedulingStageToContentName = {
+  schedulingCoaching: `scheduleSession`,
+  schedulingTherapy: `scheduleSession`,
+  schedulingBoth: `scheduleSession`,
+  referral: `referralWelcome`
+}
+// future optimization: memoize return values and eschew redunant airtable callouts
```
craigspaeth: 👍 neat idea

apps/jobs/lib/signupStage-notifications.js
```diff
@@ -0,0 +1,159 @@
...
+
+/**
+ * Finds leads in Airtable and sends them an SMS + email notification from
+ * the copy also managed in Airtable.
+ */
+export default async () => {
+  if (isNotificationPeriod()) {
+    const leads = await at.find({
+      table: LEADS_TABLE,
+      filter: 'isNotified=0'
```
craigspaeth: Nice!

apps/jobs/lib/signupStage-notifications.js
```diff
@@ -0,0 +1,159 @@
...
+
+        const emailRes = await sendEmail({
+          email: lead['emailAddress'],
+          signupStage,
+          variables
+        })
+
+        emailRes.id
+          ? updateReceivedNotifiationsColumn({ lead })
+          : console.log(
```
craigspaeth: Nit-pick: `console.error` would be slightly more appropriate as that'll send data to stderr—which has [minor benefits](https://stackoverflow.com/questions/25532778/node-js-console-log-vs-console-info) wrt tools like Papertrail or other unix processes. Something we can address separately with a logging ticket though.

apps/jobs/lib/signupStage-notifications.js
```diff
@@ -0,0 +1,159 @@
...
+    }
+  })
+  console.log(
+    `AT record update ${res.id ? `success` : `fail`} for Lead id: "${
+      lead.id
+    }", signupStage notification: "${lead.signupStage}"`
+  )
+}
+
+const isNotificationPeriod = () => {
```
craigspaeth: 👍 nice function and use of moment-timezone here

craigspaeth:
LGTM! Just nit-picks in the comments. I'll have to take a stab at fixing those tests.

PR #144, opened 2018-08-30
Email templates

lionsdwarf:
@craigspaeth This PR contains infrastructure modification for react email templating. Template markup to follow in subsequent PR.


apps/jobs/email-templates/index.js
```diff
@@ -0,0 +1,24 @@
...
+      </Item>
+      <Item>
+        <Box
+          cellSpacing={20}
+          width='100%'
+          style={{ borderTop: '3px solid black' }}
+        >
+          <Item>
+            <Span color='gray' lineHeight={20}>
+              {body}
```
craigspaeth: So far so good! One thing to note about this (that you probably already have in mind) is in the following PR this will need to get smarter than just injecting the HTML of the markdown in here. We'll want to parse the markdown [into a data structure](https://github.com/findOctave/cortex/blob/master/apps/api/models/content-module.js#L91) we can inject granularly.

For instance, the [transactional emails](https://app.zeplin.io/project/5b33fcd40b34d9ec342762ac/screen/5b7cfedb2cce4a21cea27b7a) will have an h1, p, and a content we want to inject, e.g.

````javascript
({ subject, content }) =>
  renderEmail(
    <Item>
      <h1>{content.h1}</h1>
      <a href={content.href[0]}>{content.a}</a>
    </Item>
  )
````

**Comment by craigspaeth** [comment]
👍 Looking good so far! After fixing the merge conflict this is good to go by me. Thanks!

File: apps/jobs/email-templates/index.js (hunk 1 above)
Comment (lionsdwarf) [reply apps/jobs/email-templates/index.js]: Copy that! The dynamic markdown is next on my to-do list.

File: apps/jobs/email-templates/index.js [hunk 2]
```diff
@@ -0,0 +1,18 @@
...
+*  Require the template file where file name = "signupStage"
+*  If no such file exists, mount the default template file
+*/
+
+export default props => {
+  let template
+  try {
+    template = require(`./react-templates/${props.templateName}`)
+  } catch (err) {
+    template = require(`./react-templates/${DEFAULT_TEMPLATE}`)
```
craigspaeth: 👍 nice

apps/jobs/email-templates/index.js
```diff
@@ -0,0 +1,18 @@
+import { renderEmail } from 'react-html-email'
+
+const DEFAULT_TEMPLATE = `default`
```
craigspaeth: I'm surprised Prettier doesn't auto-fix this to use single quotes because you're not using template string features here 🤷‍♂️ —if the linter doesn't correct it then it's okay by me, but just FYI I tend to use single quotes `'` in this case if we want to keep consistency.

apps/jobs/email-templates/react-templates/default.js
```diff
@@ -0,0 +1,29 @@
+import { Email, Item, Span, A, Image } from 'react-html-email'
```
craigspaeth: Such a cool lib!

apps/jobs/email-templates/react-templates/schedulingBoth.js
```diff
@@ -0,0 +1,29 @@
+import { Email, Item, Span, A, Image } from 'react-html-email'
```
craigspaeth: Nit-pick: 🤔 there should be a prettier/linter for file-naming conventions. I'd prefer we keep consistent with using kebab case for the file name here `scheduling-both.js`. Not a blocker though, just FYI.

apps/jobs/lib/signup-stage-notifications.js
```diff
@@ -57,18 +58,10 @@ export const toTwilioVars = (lead, sms) => {
...
-    schedulingBoth: 'scheduleSession',
-    referral: 'referralWelcome'
-  }[lead.signupStage] || lead.signupStage)
-
-const notificationContent = async lead => {
+const fetchNotificationContent = async lead => {
   const records = await at.find({
     table: 'notifications',
-    filter: `name = '${notificationName(lead)}'`
+    filter: `name = '${lead.signupStage}'`
```
craigspaeth: 👍 nice

apps/jobs/lib/signup-stage-notifications.js
```diff
@@ -87,7 +80,16 @@ const sendSMS = async (lead, content) => {
 }
 
 const sendEmail = async (lead, content) => {
-  const html = injectVariables(lead, marked.parse(content.emailBody || ''))
+  const emailContentModule = await fetchEmailBodyContentModule(
+    content.contentModule[0]
+  )
+  content.emailBody = emailContentModule.copy
+  const compiledBody = injectVariables(lead, content.emailBody)
+  const body = await airtableToModel({
```
craigspaeth: 👍 nice

apps/jobs/test/signup-stage-notifications.test.js
```diff
@@ -106,17 +99,27 @@ test('main sends notifications email, sms, for signup stage', async () => {
...
       {
         name: 'scheduleSession',
         sms: 'Hello',
         emailSubject: 'Hi',
-        emailBody: 'Hello World'
+        emailBody: 'Hello World',
+        contentModule: ['xxxx']
       }
     ])
+  at.findOne.mockResolvedValueOnce({
```
craigspaeth: 👍 nice update to the test.

craigspaeth:
LGTM! Only non-blocking nit-pick comments. Thanks for the updates. Excited to merge this!

apps/jobs/email-templates/index.js
lionsdwarf: This reason is not a great one, but I tend toward the template string because of the way it syntax highlights in my text editor. I'm happy to stick to single quotes for consistency going fwd.

apps/jobs/email-templates/react-templates/schedulingBoth.js
lionsdwarf: Dang it!! I reverted to an old bad habit here. 

I am still working on this file and will rename it for subsequent PR. @craigspaeth

PR #147, opened 2018-08-30
Marketing secondary pages

mkandler:
- Adds `/therapy` `/coaching` `/classes` pages
- Renames HomeOverview to IntroBlock to be reused
- Hardcoded Header (seems better with link highlighting and everything)
- Adds `childModules` to `ContentModule` type


apps/api/test/models/content-modules.test.js
```diff
@@ -4,22 +4,24 @@ jest.mock('at', () => ({
   base: jest.fn().mockReturnValue({ create: () => {} })
 }))
 
-test('airtableToModel transforms markdown copy in airtable to json', () => {
-  const conentModule = ContentModule.airtableToModel({ copy: '# Hello World' })
+test('airtableToModel transforms markdown copy in airtable to json', async () => {
+  const conentModule = await ContentModule.airtableToModel({
```
craigspaeth: A nice benefit of using that "thunk" function thing I mention above is that it allows the parent function to remain ["pure"](https://medium.com/javascript-scene/master-the-javascript-interview-what-is-a-pure-function-d1c076bec976) which, among a couple benefits, keeps functions easy to test and helps contain the [red/blue function problem](http://journal.stuffwithstuff.com/2015/02/01/what-color-is-your-function/) by keeping it from being `async`.

apps/marketing/components/ListsBlock/index.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import classNames from 'classnames/bind'
+import styles from './lists-block.css'
+
+const cx = classNames.bind(styles)
+
+const List = ({ h1, li }) => (
```
craigspaeth: 👌 UI code looking 💯 as usual!

craigspaeth:
Looks great! I added some jargon-ey FYI comments, but not blockers by any means—if you feel inspired feel free to follow up with a PR addressing it. Good stuff, thanks!

PR #157, opened 2018-09-05
leads model update to mirror picklist option removals in airtable

lionsdwarf:
@craigspaeth Keys for signupStageMap now reflect the revisions made to the signupStage picklist options in the Leads airtable. Was not sure how/where the map values are used, so I did not change "ONBOARDING" to "APPBOARDING". If this should be changed, please advise and I will update. Thanks!


apps/api/models/lead.js
```diff
@@ -31,12 +31,10 @@ const fields = (method = 'read') =>
 
 const signupStageMap = {
   referral: 'REFERRAL',
-  consulting: 'CONSULTING',
   schedulingCoaching: 'SCHEDULING_COACHING',
   schedulingTherapy: 'SCHEDULING_THERAPY',
   schedulingBoth: 'SCHEDULING_BOTH',
-  onboarding: 'ONBOARDING',
-  onboarded: 'ONBOARDED'
+  appboarding: 'ONBOARDING'
```
craigspaeth: Le'ts change the value to APPBOARDING too—onboarding was an old term I used to describe onboarding to the "app", but it got confusing with onboarding to therapy/coaching/classes and I just didn't get around to updating this.

PR #158, opened 2018-09-05
Add tests to charge sessions

craigspaeth:
A little bit of cleanup on the charge job. I tested it manually and it seems to be working...

image

Keep in mind there's a couple funky conditions that need to line up to be charged:

1. The client must complete [scheduling their first sesion](https://app.zeplin.io/project/5b33fcd40b34d9ec342762ac?seid=5b7dcd1b79c72e4a1789b66e) which creates an Auth0 account that links together the Stripe customer and Cliniko patient
2. The appointment in Cliniko needs to be marked "Arrived"
3. The appointment must be at least 48 hours old
4. The appointment has a billable item associated to it and a charge doesn't already exist in Stripe


craigspaeth:
Did an `npm update` for good measure too—hence the big diff.

PR #164, opened 2018-09-07
Repro-prod-locally and fixed prod. CSS issue

craigspaeth:
Phew! Lots of noise here, but out of it comes two important things:

1. Able to run `npm run prod` locally to reproduce a production-like environment with the app "built"
2. A small CSS issue where I had to wrap `lib/styles/reset` in `:global` for it to work, fixing the prod issue that started me down this rabbit hole


craigspaeth:
Noooooo! It's still happening 😱 😿 

image

PR #167, opened 2018-09-11
Email templates

lionsdwarf:


apps/jobs/email-templates/react-templates/default.js
```diff
@@ -1,29 +1,189 @@
...
+import { Email, Box, Item, A, Image } from 'react-html-email'
 
+const { APP_URL } = process.env
+const SUPPORT_EMAIL = 'support@findoctave.com'
+
+// style constants
 const IMG_WIDTH = 600
 const IMG_HEIGHT = 337
+const CONTENT_WIDTH = '472px'
+const CHARCOAL_COLOR = '#232323'
```
craigspaeth: I wonder if we could import the CSS module color variables here.

@mkandler this is an example of where I like CSS-in-JS solutions b/c it's more straightforward to share code in these edge cases like templating emails server-side. That said, I'm pretty sure there _is_ a way to do this with CSS modules—just might require jumping through some hoops for server-side webpack or babel config to support `import foo from 'foo.css'`.

apps/jobs/email-templates/react-templates/default.js
```diff
@@ -1,29 +1,189 @@
...
-    </Item>
-    <Item>
-      <Span fontSize={15}>{h4[0]}</Span>
-    </Item>
-    <Item align={`center`}>
-      <A href={href[0]} textDecoration={`none`}>
-        {a[0]}
-      </A>
-    </Item>
+  <Email title={`Octave`} headCSS={stylesheet}>
```
craigspaeth: Seeing React code for email templating puts a smile on my face even though I know this library hasn't been the smoothest to use 😸

PR #194, opened 2018-09-25
Redundant code cleanup

lionsdwarf:
@craigspaeth, contained within are modifications to replace .then/.catch with try/catch and eliminate redundant code, [as suggested] (https://github.com/findOctave/cortex/pull/176).

As mentioned in Slack, the Cliniko post request for Assessment Note is failing with a message-less 403. This is happening on master branch as well. I suspect upstream data config is to blame, but not troubleshooting now with so much Cliniko tinkering going on...

screen shot 2018-09-25 at 2 34 00 pm


apps/jobs/lib/sync-assessments.js
```diff
@@ -85,54 +84,55 @@ export const buildAssessmentNoteData = (record, patientsEmailToId) => ({
...
         [CLINIKO_NOTE_ID_FIELD_NAME]: parseClinikoNoteIdFromLink(res.links.self)
       }
     })
-    .catch(err =>
-      logErr({
-        error: `Airtable record update failure: Assessement "ClinikoNoteId" update`,
-        metadata: err
-      })
-    )
+  } catch (err) {
```
craigspaeth: These `catch`s and `logErr`s above are also probably unnecessary because these `err` objects will just bubble up to the [cli `.catch(log.error)`](https://github.com/findOctave/cortex/blob/master/apps/jobs/lib/cli.js#L7)

craigspaeth:
LGTM! Left a small note about letting errors bubble up even further, but feel free to ignore it and merge if you like wrapping in your own messages for log clarity.

apps/jobs/lib/sync-assessments.js
lionsdwarf: Ah, got it. Copy that. Fixing now.

PR #198, opened 2018-09-28
Stripe components server side

lionsdwarf:


apps/portal/components/schedule-billing/index.js
```diff
@@ -1,32 +1,23 @@
...
-      <StripeCheckout token={onStripeToken} stripeKey={STRIPE_PUBLISHABLE}>
-        <Button onClick={onOpenStripe} disabled={submittedCard}>
-          {!submittedCard && (
-            <img className={cx('cc-icon')} src='static/icon-cc.svg' />
-          )}
-          {submittedCard ? 'Thanks!' : content.a[0]}
-        </Button>
-      </StripeCheckout>
+      <CreditCardForm {...creditCardForm} />
+      <ValidateCoupon {...applyCouponForm} />
```
craigspaeth: Nice breaking out of these two components!

craigspaeth:
Code mostly LGTM!

Not a blocker, but as I look this over I notice this is starting to put a lot of "business logic" in the schedule session component around creating a client > applying a coupon > handling error messages. I wonder if it would be better to move this downstream into GraphQL so that the client-side logic is something like...

````javsacript
try {
  await gql.request(`
    mutation {
      createClientFromLead(
        leadId: "${this.props.leadId}"
        stripeSource: "${stripeData.token.id}"
        couponCode: "${this.refs.couponInput.value}"
      ) { name }
    }
  `)
} catch (err) {
  this.setState({ err })
}
````

and the backend mutation resolver for `couponCode` would essentially do what the `applyCouponToLead` mutation does.

Not sure how this would play out in practice and mostly just speculating here—so like I said code LGTM and not-blocking for it.

👏 good stuff!

lionsdwarf:
@craigspaeth You're correct that the error logic could definitely live within gql.

However, I built this logic into the client bc I feel it's a client-specific implementation to swallow errors for duplicate coupon codes. As is, we are treating a dupe coupon as a success, but I can imagine a future enhancement where we notify the user that they've already applied a given coupon. As is, in this case we would just need to update the app logic. But if we build this into gql, we would probably have to change it it both places, plus any other clients relying on the mutation logic.

PR #199, opened 2018-09-28
zendesk widget added to portal app

lionsdwarf:
@craigspaeth I've generically added the zendesk widget to apps/portal/pages/_document, meaning it appears everywhere in the portal app. If we need more specificity on this, let me know and I can modify. Thanks!


apps/portal/pages/_document.js
```diff
@@ -13,6 +13,10 @@ export default class Doc extends Document {
             rel='stylesheet'
           />
           <link rel='stylesheet' href='/portal/_next/static/style.css' />
+          <script
+            id='ze-snippet'
```
craigspaeth: Easy peasy! Thanks!

PR #202, opened 2018-10-02
Format assessment notes better (+ some refactoring)

craigspaeth:
This uses Cliniko's paragraph type along with their [permitted br tags](https://github.com/redguava/cliniko-api/blob/master/sections/treatment_notes.md#permitted-html-in-paragraph-answers) to provide some basic formatting to the treatment notes.

screen shot 2018-10-02 at 3 42 13 pm

Some side notes:
- I refactored things a bit to let the `await`s bubble up error messages
- Use the log lib to get error coloring + emojis in the terminal
- Removed storing the airtableId on the Cliniko-side seeing as they weren't serving much of a functional purpose and to make the notes even less scary/data-looking
- I had to check the IT account in Cliniko as a "practitioner" and remove it from online bookings (I think this was an issue you ran into earlier)


lionsdwarf:
Very nice, @craigspaeth ! Nice find on the Cliniko support for `<br>` tags. Did you happen to try using `<div>`s ? I believe that `div`s are stripped even though the docs claim otherwise.

PR #204, opened 2018-10-02
Support grandfathering flow

craigspaeth:
This adds support for creating a lead via query params in the schedule session flow. This is so we can support "grandfathering" existing patients into Octave by letting therapists personally send their old patients a url to schedule therapy.


craigspaeth:
Going to merge b/c upcoming PRs are going to get noisy with this.

apps/portal/pages/schedule-session.js
```diff
@@ -140,9 +138,10 @@ class ScheduleSessionPage extends React.Component {
   }
 
   nextStep = async () => {
-    await this.isProceedValid() && this.setState({
-      step: this.state.step + 1
-    })
+    ;(await this.isProceedValid()) &&
```
lionsdwarf: @craigspaeth What's the deal with this semi-colon?

PR #205, opened 2018-10-03
Add Apercu

craigspaeth:
This hosts Apercu by simply putting it in the portal app's static folder and putting the @font-face tag in the reset.css. Ideally this would be more modularly used across apps—for which I tried [symlinks](https://stackoverflow.com/questions/10618418/how-can-i-implement-virtual-directories-with-node-js-and-express)—but it seemed to not be working. I'm not even sure that's the most elegant solution to share static assets anyways, so I just went for the simplest solution.


lib/styles/fonts.css
```diff
@@ -0,0 +1,14 @@
+/*
+  Font files are located in apps/portal.
+  TODO: Come up with a better place/way to share font files across apps
+*/
+@font-face {
+  font-family: 'Apercu';
```
craigspaeth: Note: I changed the `font-family` here from what was provided in the the .zip I sent you to line up with [what was referenced in the styles lib](https://github.com/findOctave/cortex/blob/master/lib/styles/index.css#L20).

lionsdwarf:
@craigspaeth Thanks for the update. I tried adding @font-face to index.css. I guess I don't understand what reset.css is for.

lib/styles/fonts.css
lionsdwarf: btw, @craigspaeth, in my attempt I had changed the index.css to use font-family 'apercu-medium-pro', in reference to the out-of-box @font-face declaration

PR #208, opened 2018-10-03
env var toggles nav classes link

lionsdwarf:


apps/marketing/components/Header/index.js
```diff
@@ -4,6 +4,8 @@ import PropTypes from 'prop-types'
 import styles from './header.css'
 
 const cx = classNames.bind(styles)
+const isActiveClasses =
+  !process.env.IS_CLASSES_ACTIVE || process.env.IS_CLASSES_ACTIVE === 'true'
```
lionsdwarf: @craigspaeth One must include IS_CLASSES_ACTIVE=false only to deactivate the link/page. In other words, default state (eg absent IS_CLASSES_ACTIVE env var) renders `/classes` link. I included the variable in our .envenc anyways just for an easy reference to the env variable name.

craigspaeth: Makes sense to me, thanks!

PR #212, opened 2018-10-04
fix for cliniko iframe validation error that prevents subsequent form…

lionsdwarf:
… inputs

@craigspaeth Do you recall the purpose of this "scrolling" attribute on the cliniko iframe? It breaks the form on mobile, as documented here: https://trello.com/c/gsbvNQ9a/44-schedule-session-flow-breaks


craigspaeth:
Hmm, no I think it was just a copy/paste fail. Thanks for fixing!

PR #216, opened 2018-10-06
Refactor sync assessments job to bypass 100 page size limit on Cliniko patients

craigspaeth:
Apologies for getting a little carried away on refactoring this job, but b/c it was pulling a limit of 100 patients from Cliniko at the beginning, the job was failing now that we have ~500 leads synced to patients in Cliniko. This instead "lazily" fetches patients 1 by 1 after finding unsynced records. My theory is this will be fine when the job run frequently b/c unsynced records shouldn't build up so fast that we hit Cliniko rate limits. The rest of the refactors just involve some naming consistency and consolidation.


lionsdwarf:
All good! Makes sense and thanks for the heads up.

PR #218, opened 2018-10-08
availability model, createAppointment enhancement to Appointment model

lionsdwarf:


apps/api/models/appointment.js
```diff
@@ -65,6 +74,14 @@ export const typeFromCliniko = apptType => ({
     })).practitioners.map(Practitioner.fromCliniko)
 })
 
+export const toCliniko = apptData => ({
+  starts_at: apptData.startTime,
+  patient_id: apptData.patientId,
+  appointment_type_id: apptData.appointmentTypeId,
+  practitioner_id: apptData.practitionerId,
+  business_id: NEW_YORK_BUSINESS_ID
```
craigspaeth: This is fine for now, but it actually won't be too much longer until we will have to face the challenge of managing Octave's SF and NY clinics in Cliniko as separate businesses.

apps/api/models/availability.js
```diff
@@ -0,0 +1,47 @@
...
+
+const fromCliniko = appts => appts.map(
+  appt => ({
+    appointmentStart: appt.appointment_start
+  })
+)
+
+export const availabilities = async (_root, args, { req }) => {
+  let appts = await cliniko.find({ 
+    url: `https://api.cliniko.com/v1/businesses/${NEW_YORK_BUSINESS_ID}/practitioners/${args.practitionerId}/appointment_types/${args.appointmentTypeId}/available_times?from=${args.startDate}&to=${args.endDate}`
```
craigspaeth: Nit picks: I'd prefer defaulting to using `const` instead of `let appts` wherever possible and it might be nicer to break up `url` into multiple lines e.g.

```javascript
const url = [
  'https://api.cliniko.com/v1',
  `/businesses/${NEW_YORK_BUSINESS_ID}`,
  `/practitioners/${args.practitionerId}`,
  `/appointment_types/${args.appointmentTypeId}/available_times`
  `?from=${args.startDate}`,
  `&to=${args.endDate}`
].join('')
const appts = await cliniko.find({ url })
```

apps/api/models/availability.js
```diff
@@ -0,0 +1,47 @@
...
+export const NEW_YORK_BUSINESS_ID = 66701
+
+export const schema = {
+  types: `
+    type Availability {
+      appointmentStart: String
+    }
+  `,
+  queries: `
+    availabilities(
```
craigspaeth: I mention this in the [tech. architecture doc](https://github.com/findOctave/cortex#tech-architecture), but I think it might be nicer to model querying through nested relationships like this more directly "along the graph" rather than accepting a bunch of args. I can foresee a bunch of root level args quickly getting hard to evolve as we need to do things like compose sending filters to Cliniko and filtering out results client-side with other local data models.

I also wonder if there's a more directly useful query we can give to the front-end instead of asking them to know things like practitioner and start/end date (which they wouldn't in the case of scheduling a consult).

So instead of this, maybe the query can look more like this....

```javascript
appointmentType(category: CONSULT) {
  availabilities {
    startAt
    endAt
  }
}
```

As I look at it I also notice, `appointmentStart` from [available times](https://github.com/redguava/cliniko-api/blob/master/sections/available_times.md) will give us an easy way to build the view of "list times slots you can click on" but we probably want to be able to show a calendar view with open blocks of time, or a `startAt` + `endAt` closer to what [availability blocks](https://github.com/redguava/cliniko-api/blob/master/sections/availability_blocks.md) returns. Technically the client-side could figure this out by combining `appointmentStart` with the duration of the appointment type—but that starts to get complicated 🤔

craigspaeth:
This is great work, but I think we've only scratched the surface of building a calendar UI with this. I think there'll be a decent amount more to figure out when moving to the front-end such as the blocks vs. appointmentStart thing I mentioned above and being able to query the data in a way that makes sense for consults (picking times across multiple practitioners) vs. therapy/coaching (choosing person ahead of time) vs. classes (🤷‍♂️ we'll see what other edge cases), and some-day in the near future there'll multiple businesses we'll have to account for in our code.

Had I the foresight to see what issues we'd run into with the iframe I'd probably have picked this route form the get-go. However, given the time constraints, as much as I don't love the Cliniko iframe I think it'd be safer to pick this work back up post-launch when we have the breathing room to build it thoughtfully.

Thanks for exploring this! We'll definitely use this, but for now, I'm going to close this PR and we can consider reviving it post-launch.

apps/api/models/availability.js
lionsdwarf: Hear you loud-n-clear regarding the nested relationships query structure--good call and thanks for highlighting / revising the approach there. Makes sense to me. 

Regarding time blocks and start/end times...For the sake of time and the goal of quickly building a scheduler that omits a Patient Create form, I took the simplest approach possible: mimic the existing calendar scheduler.  This approach is built on the available_times API endpoint, which unfortunately returns only appointment start_time values. 

Without a doubt, a more elaborate scheduler with time blocks will have broader considerations. My initial research indicates we will likely have to synthesize data from a number of different cliniko endpoints in order to generate the basis for the UI.

craigspaeth: Make sense, it'll be nice to write a clean API on top of all of those endpoints in Cliniko down the line.

PR #222, opened 2018-10-09
Lots o' refactors

craigspaeth:
Apologies for the large diff. This brings a number of refactors, bug fixes, and tests. I'll comment more in-line.


apps/api/models/appointment.js
```diff
@@ -82,7 +83,21 @@ export const fromCliniko = appt => ({
   lead: async () =>
     Lead.findOrCreateFromClinikoPatient(
       await cliniko.find({ url: appt.patient.links.self })
+    ),
+  leads: async args => {
```
craigspaeth: For classes (or group appointments) we need to model correctly the fact that an appointment can be related to multiple leads (attendees).

apps/api/models/appointment.js
```diff
@@ -126,13 +141,35 @@ export const appointmentTypes = async _root => {
...
+
 export const appointments = async (_root, args, { req }) => {
-  let appts = await cliniko.find({ resource: 'individual_appointments' })
-  if (args.bookingIpAddress) {
-    appts = appts.filter(
-      appt => appt.booking_ip_address === args.bookingIpAddress
+  let appts = _.flatten(
+    await Promise.all([
+      cliniko.find({ resource: 'individual_appointments' }),
+      cliniko.find({ resource: 'group_appointments' })
```
craigspaeth: We'll consider appointment one data model from our perspective, so we combine them here which makes searching across them via IP address simple for use on the front end.

apps/portal/components/schedule-billing/index.js
```diff
@@ -1,10 +1,17 @@
...
-import styles from './schedule-session-billing.css'
+import CreditCardForm from '../credit-card-form'
+import React from 'react'
+import styles from './schedule-billing.css'
 
 const cx = classNames.bind(styles)
-
+const gql = new GraphQLClient(process.env.APP_URL + '/api')
+const STRIPE_PUBLISHABLE = process.env.STRIPE_PUBLISHABLE
+const getCouponErrorMessage = err =>
```
craigspaeth: I moved all of this Stripe elements stuff inside the schedule billing component which makes it easier to use in class scheduling.

apps/portal/components/schedule-billing/index.js
```diff
@@ -19,27 +26,166 @@ const stripeInjectedPTag = (p, a, href) => {
...
+      applyCouponForm: {
+        couponCode: '',
+        handleUserInput: this.handleCouponFormInput
+      }
+    }
+  }
+
+  async componentDidUpdate (prevProps) {
+    if (this.props.submittingBilling && !prevProps.submittingBilling) {
+      if (await this.isProceedValid()) this.props.onSuccess()
```
craigspaeth: Here's where I get a little weird with sharing state across components and realize that this maybe deserves a more aggressive refactor with Redux or something.

What's going on here is...
- The sibling `ScheduleFooter` component calls `onNextStep` which changes the state of the root component `submittingBilling = true`
- This signals back down to this billing component that it's time to validate and submit
- If successful it signals back up to the root component via `onSuccess` that it can proceed to the next step or via `onError` that it needs to stay put

apps/portal/components/schedule-billing/schedule-billing.css
```diff
@@ -1,7 +1,7 @@
 @import 'styles';
 
 .intro {
-  @apply --esteban-title;
+  @apply --esteban-subheader;
```
craigspaeth: Bump down a size b/c Meg's copy is long

apps/portal/components/schedule-session-choose/index.js
```diff
@@ -4,28 +4,17 @@ import styles from './schedule-session-choose.css'
...
-                practitionerPortraitFilenamesToURLs[clinician.portraitFilename]
-              }
-            />
-          )}
+          {clinician.portraitUrl && <img src={clinician.portraitUrl} />}
           <h3>{clinician.name}</h3>
           <h4>${clinician.rate} per session</h4>
           <p>{clinician.description}</p>
-          <Button variant='medium' onClick={() => chooseClinician(clinician)}>
+          <Button variant='medium' onClick={() => onChooseClinician(clinician)}>
```
craigspaeth: Trying to be consistent about naming these callbacks `on`Something

apps/portal/lib/appointment.js
```diff
@@ -1,35 +1,68 @@
...
 import { GraphQLClient } from 'graphql-request'
 import publicIp from 'public-ip'
 import moment from 'moment'
+import { ACCEPT_TERMS_MSG } from './client'
 
 const gql = new GraphQLClient(process.env.APP_URL + '/api', { headers: {} })
 
-export const justBookedForCategory = async category => {
+export const confirmJustBooked = async ({ category, setState, state }) => {
+  setState({ disabledMsg: 'Confirming appointment...' })
```
craigspaeth: This is a little funny, but what I'm doing here is extracting state transitions and the corresponding "business logic" into our "view model" libs.

In a way, it's achieving what something like Redux encourages (if you squint your eyes the injected `setState` is like `dispatch(action)` and pure functions like `confirmFoundAndNextStepOrRejected` are like reducers.

That said, I acknowledge this is pretty clunky and this probably warrants some more traditional state management like Redux, Apollo Client, or something with React context. For now, this gets us some DRY-er code and test coverage on state transitions,.

apps/portal/lib/practitioner.js
```diff
@@ -77,12 +83,20 @@ export const buildCards = arr =>
     .flatten()
     .value()
 
-export const cardsFromApptTypes = appointmentTypes => {
+export const mapPortraitFilenamesToURLs = images =>
```
craigspaeth: Moved the portrait mapping inside this practitioner lib to better encapsulate it with this other practitioner data logic.

apps/portal/components/schedule-billing/index.js
lionsdwarf: Good call. I much prefer this design approach, which promotes separation of concerns. I recognize the conveniences of the previous pattern, eg bubbling up events / state to a fat parent component, but such approaches often intermingle domains that should remain separate. Nice move!

apps/portal/components/schedule-billing/index.js
lionsdwarf: _queue previous comment's mention of "the conveniences of the previous pattern"_ I'm following you!

apps/portal/pages/schedule-class.js
```diff
@@ -41,51 +42,59 @@ export default class ScheduleClassPage extends React.Component {
...
-      step: this.state.step + 1,
-      disabledMsg:
-        {
-          1: 'Please fill out billing info first'
-        }[this.state.step] || null
-    })
+  onNextStep = () => {
+    if (this.state.step === 0) {
+      return Appointment.confirmJustBooked({
+        category: 'CLASS',
```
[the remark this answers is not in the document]
lionsdwarf: Sexy.

PR #223, opened 2018-10-10
Add transfer client flow

craigspaeth:
This adds the "grandfathering" flow where we can invite patients transferred from existing therapy practices to enter our systems with a URL that includes name/email/phone. It's then up to their therapist to transfer their appointments to Cliniko.

cap


lionsdwarf:
LGTM!

PR #225, opened 2018-10-10
styles for assessments zero state

lionsdwarf:
Took me a bit to reverse engineer the proper url, but this is working now. Do we want to add a border to the nav bar?


lionsdwarf:
screen shot 2018-10-10 at 8 48 03 am
screen shot 2018-10-10 at 8 47 43 am

PR #243, opened 2018-10-11
Line elements

mkandler:
- Adds in the line elements (animating with scroll)
- Adds icons to the brown bubbles
- Including some missing images


apps/api/models/content-module.js
```diff
@@ -132,6 +132,7 @@ export const airtableToModel = record => {
     return { ...acc, [tag]: _.compact([...(acc[tag] || []), _.trim(text)]) }
   }, {})
   const images = args => {
+    if (!record.images) return null
```
craigspaeth: Good catch!

craigspaeth:
LGTM!

PR #244, opened 2018-10-11
Support markdown inside paragraph tags

craigspaeth:
This renders the full HTML inside paragraph tags so we can support things like:

```
# Billing information

We use [Stripe](https://stripe.com/about) to authorize payments, so rest assured.
```

image


apps/portal/components/schedule-billing/index.js
```diff
@@ -142,19 +129,15 @@ export class ScheduleBilling extends React.Component {
...
+          <p dangerouslySetInnerHTML={{ __html: this.props.content.p[2] }} />
           <CreditCardForm {...this.state.creditCardForm} />
           <ApplyCouponForm {...this.state.applyCouponForm} />
         </div>
         <div className={cx('better')}>
           <img src={this.props.content.images[0].url} />
           <div className={cx('copy')}>
             <h2>{this.props.content.h2}</h2>
-            <p>{this.props.content.p[3]}</p>
+            <p dangerouslySetInnerHTML={{ __html: this.props.content.p[3] }} />
```
craigspaeth: Make sure to use this `dangerouslySetInnerHTML` in these cases.

lionsdwarf:
Nice! Copy that.

apps/api/models/content-module.js
```diff
@@ -91,45 +91,51 @@ export const schema = {
 }
 
 export const airtableToModel = record => {
-  const $ = cheerio.load(marked.parse(record.copy || ''))
+  const $ = cheerio.load(
+    marked.parse((record.copy || '').replace(/\n/g, '\n\n'))
```
craigspaeth: Sometimes marked will consider text separated by newlines a part of the same element:

```
This sentence
[Some link]()
```

outputs `<p>This sentence <a>Some link</a></p>. This is pretty counter-intuitive to the way we're using the root-level markdown elements as structure, so we're replacing single newlines with two newlines to remove confusing before parsing.

apps/api/models/content-module.js
```diff
@@ -91,45 +91,51 @@ export const schema = {
...
+      $('body > p')
+        .map((i, el) => {
+          const hasTextNodes = !!$(el)
+            .clone()
+            .children()
+            .remove()
+            .end()
+            .text()
+          const els = []
+          if (hasTextNodes) {
```
craigspaeth: Should have extracted this to it's own function and put some comments here, but this block of basically does a couple things:

1. If there's text nodes it injects those into `p`s
2. If there's no text notes (e.g. `<p><a>Foo</a></p>`) then it injects that into `a` or the corresponding tag
3. If it's an `a` then it plucks out the `href`s

apps/api/models/content-module.js
mkandler: Is this going to replace all `\n` with `\n\n` though? I guess that's ok?

apps/api/models/content-module.js
```diff
@@ -91,45 +91,51 @@ export const schema = {
...
-        .children()
-        .remove()
-        .end()
-        .text()
-      return { tag: $(el).get(0).tagName, text }
-    })
-    .get()
+  const pEls = _.flatten(
+    _.compact(
+      $('body > p')
```
mkandler: Could you also add the same functionality for `blockquote`? There are a few cases where I need to use them with anchor tags for those `<Aside>` elements

apps/api/models/content-module.js
craigspaeth: Yeah, good point. However, I think in most cases where someone thinks they want newlines they actually want separate paragraphs or `<br>`s (which are also fairly counter intuitive in markdown with four spaces at the end of the line). Given the way we're using markdown for data structure, I'd rather err on the side of intuitively supporting the rigid elements in place vs. supporting the full range of markdown functionality (this allowing inline `p` markdown being an exception :-P)

craigspaeth:
@mkandler I added blockquote support now too

mkandler:
Nice! LGTM ✨

PR #248, opened 2018-10-11
Don't HTML encode characters if there's no sub-elements

craigspaeth:
Fixes this:

image


apps/api/models/content-module.js
```diff
@@ -90,17 +90,43 @@ export const schema = {
...
     .remove()
     .end()
     .text()
+  const hasTextNodes = !!strippedText
+  const onlyText = strippedText === $el.text()
   const els = []
   // If there's text nodes inject into `p` field
   if (hasTextNodes) {
-    els.push({ tag, text: $el.html() })
+    els.push({ tag, text: onlyText ? $el.text() : $el.html() })
```
craigspaeth: This might be getting too clever, but basically if there are no sub-elements in the paragraph/blockquote we use `.text()` to return the plain text.

/me starting to regret this DIY Airtable CMS 🙃

mkandler: I think it's a good clever! We only really ever want HTML output if there are sub-elements present anyway. (At least in all the cases I can think of)

Too late to turn back now!

PR #252, opened 2018-10-12
Metatags editable in airtable

mkandler:
All pages have editable Meta Tags in Airtable with the format:
`metaData{{pageName}}`


craigspaeth:
Nice!

PR #269, opened 2018-10-15
Add id to class list so we can use an hash link

craigspaeth:
Which means we can use a simple hash link 
image


craigspaeth:
cc @mkandler

PR #276, opened 2018-10-16
Post launch qa

mkandler:
- Creates a disabled state for buttons without `url` or `onClick`
- Buttons scroll to section if the url starts with `#` (NOTE: this `#class-list` preferred over `/classes#classlist` for anchors on the page)
- Learn More in hero scrolls down the page


apps/marketing/components/Button/index.js
```diff
@@ -1,15 +1,47 @@
...
+    if (url && url.indexOf('#') === 0) {
+      e.preventDefault()
+      const id = url.split('#')[1]
+      const el = document.getElementById(id)
+      window.scroll({
+        behavior: 'smooth',
+        left: 0,
+        top: el.getBoundingClientRect().top - 100
+      })
+      history.pushState(null, null, `#${id}`)
```
craigspaeth: 👍 nice, I was about to comment on Next's `Router.push` but then I realized we don't want to use the router b/c it would unnecessarily trigger `getInitialProps` and such. LGTM!

PR #278, opened 2018-10-17
Auth over querying leads by id

craigspaeth:
The gets authorization (not authentication, which comes when we build logged-in mode) rolling by using [jwt](https://jwt.io/) based authorization in our GraphQL API. In this use case, we're adding authorization to our `lead(id: X)` query and generating the JWT in our email job. I'll comment more inline.


apps/api/lib/jwt-context.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * Apollo server middleware that verifies a JWT and attaches is a ctx.user
+ */
+import jwt from 'jsonwebtoken'
+
+export default secret => ({ req }) => {
```
craigspaeth: If a JWT is added to the `Authorization: Bearer {Token}` header of a request, we can validate it was generated from a trusted source (aka our servers in the email job below) upfront before we even begin to resolve the GraphQL query.

apps/api/lib/jwt-context.js
```diff
@@ -0,0 +1,11 @@
...
+  const token =
+    req.headers.authorization && req.headers.authorization.split(/^Bearer /)[1]
+  const user = token && jwt.verify(token, secret)
```
craigspaeth: If a JWT is added to the `Authorization: Bearer {Token}` header of a request, we can validate it was generated from a trusted source (aka our servers in the email job below) upfront before we even begin to resolve the GraphQL query.

apps/api/models/lead.js
```diff
@@ -135,7 +135,14 @@ export const findOrCreateFromClinikoPatient = async patient => {
   return fromAirtable(lead)
 }
 
-export const lead = async (_root, args) => {
+export const authorizeLeadQuery = (args, ctx) => {
+  if (args.id && (!ctx.user || ctx.user.id !== args.id)) {
```
craigspaeth: At this point the JWT is validated and parsed, so we can trust the content of its payload, which is injected in `ctx.user`. In this case, we want to prevent people from being able to query for leads who are not themselves (we have yet to implement admin roles), so we check if the `id` in args is the same as the `id` provided by the payload of the JWT.

apps/portal/pages/schedule-session.js
```diff
@@ -20,6 +18,11 @@ export default class ScheduleSessionPage extends React.Component {
   }
 
   static async getInitialProps ({ query }) {
+    const gql = new GraphQLClient(process.env.APP_URL + '/api', {
+      headers: {
+        authorization: `Bearer ${query.token}`
```
craigspaeth: In the schedule session flow we pass the generated JWT token from the URL into our GraphQL query where the above auth logic runs.

apps/jobs/lib/signup-stage-notifications/index.js
```diff
@@ -4,20 +4,27 @@
...
+const { EMAIL_FROM_ADDRESS, TWILIO_SERVICE_SID } = process.env
 
-export const injectVariables = (lead, copy) => {
+export const injectVariables = (
+  lead,
+  copy,
+  { JWT_SECRET, APP_URL } = process.env
+) => {
+  const payload = { role: 'lead', id: lead.id }
+  const token = jwt.sign(payload, JWT_SECRET, { expiresIn: '30 days' })
```
craigspaeth: In the "We welcome you to schedule therapy" email job we generate a minimal JWT that is authorized as the `role: 'lead'` and contains the `id` of the lead we're emailing. That is then encrypted, or "signed", with our `JWT_SECRET` so our server can `verify` it later as a trusted token. We set an `expiresIn` because we don't want these to live too long in case someone gets a hold of this link (we could wind this down to an even shorter-lived time if we want to be more careful).

craigspaeth:
Going to merge b/c I have more like this coming, but plz feel free to comment.

PR #279, opened 2018-10-17
Authorize lead(email: X) queries to only return id + email

craigspaeth:
The way we need to anonymously pull a lead from the backend of Cliniko is not great for security—check another reason why we need to replace this iframe business with our own UI. In the meantime, this adds some authorization logic that denies querying a lead by email and asking for more than the id/email (with `id` being the only leak-able info in that query, which is okay after the last auth PR). Unfortunately, this slightly affects the UX—we can't inject the phone number in the confirmation screen and have to use generic copy "a care coordinator will call you at the phone number you provided earlier"—but I think that's an okay tradeoff to make sure we're not leaking data we shouldn't be.


apps/api/models/lead.js
```diff
@@ -135,14 +135,21 @@ export const findOrCreateFromClinikoPatient = async patient => {
   return fromAirtable(lead)
 }
 
-export const authorizeLeadQuery = (args, ctx) => {
-  if (args.id && (!ctx.user || ctx.user.id !== args.id)) {
-    throw Error('Not authorized')
-  }
+export const onlyEmailOrIdFields = ast => {
```
craigspaeth: This is a bit ugly, but what it's doing is traversing the AST (abstract syntax tree) of the GraphQL query and checking if only the `email` and `id` fields are included. [Relevant thread](https://github.com/graphql/graphql-js/issues/19)

craigspaeth:
Again, going to merge to try and make sure this is deployed before we start getting _real_ traffic.

PR #283, opened 2018-10-19
Billing group appts

lionsdwarf:


apps/jobs/lib/charge-sessions.js
```diff
@@ -1,5 +1,5 @@
 /**
- * A job that finds concluded coach/therapy sessions a charges the client for
+ * A job that finds concluded coach/therapy/group sessions and charges the client for
```
craigspaeth: Non-blocking nitpick: From our "domain" we'd call group appointments "classes" (unless we start offering group therapy as a different thing). _Naming Things_ 🤔

apps/jobs/lib/charge-sessions.js
```diff
@@ -9,66 +9,46 @@ import * as at from 'at'
 import * as log from './log'
 import moment from 'moment'
 
-export const isValidAppt = appt =>
-  Boolean(appt.practitioner && appt.patient && appt.appointment_type)
+/** ** Validation Methods ****/
```
craigspaeth: Non-blocking nitpicks: 
- "Methods" technically are functions on objects in the object-oriented sense, so validation "functions" might be more appropriate... but to-may-to, to-mah-to 🍅
- When I start using header comments to organize a large file like the `/** ** X ** **/` here I find that's a sign that the module should be broken up into multiple files e.g. a `jobs/lib/charge-sessions/validation.js` and `jobs/lib/charge-sessions/formatting.js`.

apps/jobs/lib/charge-sessions.js
```diff
@@ -125,44 +139,155 @@ export const chargeFromAggregate = (
...
+    })
+  )
+}
+
+/**
+ *  Given a list of Cliniko Group Appointments, get the Attendee join records and
+ *  format each GroupAppointment record so that it contains the necessary
+ *  Appointment fields
+ *
+ *  Schemas:
```
craigspaeth: 👍 nice comments here.

apps/jobs/test/charge-sessions.test.js
```diff
@@ -11,11 +11,130 @@ jest.mock('at', () => ({
...
+      patient_arrived: true,
+      did_not_arrive: false
+    },
+    historicalCharges: []
+  }
+  const isChargableAppt = cs.isChargableAppt(aggregate, [])
+  expect(isChargableAppt).toEqual(false)
+})
+
+test('isChargableAppt returns false for unbillable appts', () => {
```
craigspaeth: ❤️ the test coverage

apps/jobs/lib/charge-sessions.js
```diff
@@ -82,16 +62,47 @@ export const buildCost = (price, dollarsDiscount) =>
...
+ *  The Cliniko Attendee joiner between Group Appointment and Patient has an "attended" field
+ *  with three possible values: true, false, and null
+ *  By contrast, the Cliniko Appointment record represents
+ *  "attended" with two properties: patient_arrived, patient_did_not_arrive
+ *
+ *  This function translates the "attended" states into the Appointment format
+ *
+ *  @param {Boolean} isArrived - boolean value for the Cliniko Attendee "attended" property
+ */
+export const conformAttendeeAttendanceToAppointmentAttendance = isArrived => {
```
craigspaeth: Non-blocking nitpick: I already expressed this on Slack but just a reminder that these function names could probably be terser and rely more on the context/arguments to make sense e.g. (contrived example) `toArrivedFieldFrom(attendee)`

craigspaeth:
👏 great stuff!! I only have small nitpicks. Overall looks good, I'm just noticing this job is getting complicated enough that it probably needs to be broken down into multiple modules—but that's for another refactor.

apps/jobs/lib/charge-sessions.js
lionsdwarf: Copy that!

apps/jobs/lib/charge-sessions.js
lionsdwarf: Good point.

PR #286, opened 2018-10-21
Redirect app

craigspaeth:
Solves this redirect request for now and in the future by loading a set of path -> url pairs from Airtable. I've found this request becomes common enough (for SEO purposes and shortcut use cases like "I want to use this findoctave.com/my-favorite-press-release url to link to my PDF") that it's worth making this editable by non-devs.

image


lionsdwarf:
Cool--LGTM @craigspaeth !!

PR #290, opened 2018-10-22
Continue to update receivedNotifications despite errors to avoid spam

craigspaeth:
This is part 1 to a couple minor job refactors to be better about error handling. This prevents us from accidentally spamming notifications in case a Twilio (e.g. missing phone number—which is a separate ticket) or Mailgun call fails by making all the SMS/email/AT-update calls parallel.

image


apps/jobs/lib/signup-stage-notifications/index.js
```diff
@@ -116,29 +113,28 @@ const sendEmail = async (lead, content) => {
...
+    updateReceivedNotifications(lead)
+  ])
 }
 
 export default async (now = moment()) => {
   if (!isNotificationPeriod(now)) return log.msg('Wrong time for email')
   const leads = await at.find({
     table: 'leads',
-    filter: "AND(isNotified=0, signupStage != '')"
+    filter: "AND(NOT(FIND(signupStage, receivedNotifications)), signupStage != '')"
```
craigspaeth: Small thing—but I embedded the isNotified formula here so we can have less-dependencies on the Airtable structure. Reason being it makes keeping staging/prod. bases in sync easier and wherever we can avoid storing "code" outside the codebase the better. This is a bit of a contrived example but if you were to think of Airtable as a database `isNotified` is similar to a [stored procedure](https://kevinlawry.wordpress.com/2012/08/07/why-i-avoid-stored-procedures-and-you-should-too/) (anti-pattern) whereas keeping this in the codebase is more like a SQL query (acceptable).

PR #311, opened 2018-11-01
Beginning of data cleanup job

craigspaeth:
This starts a data cleanup job intended to run on 10 minute intervals. Like I mentioned over Slack, Airtable doesn't have last modified date, so this is a compromise to the AP-database-like node-syncing theory I ranted about over Slack.

Instead, this does something very simple for now that we can expand on as needed. It finds patients updated in the last 15 minutes, then queries Airtable's lead records, and reduces a current "truth" for shard lead fields (name, email, phone). It then syncs back out to both systems that truth.

The rules for reducing this "truth" is very simple right now: it basically merges the name/email/phone of Airtable/Cliniko—preferring Cliniko, and doesn't leave blank fields. Eventually, this can be expanded to more services, e.g. syncing to Mailchimp lists—or more advanced rules like checking the timestamp (when Airtable supports it) to pick the fields from the last written record.

Side note: This code does end up importing from the API app quite a bit, so I think eventually these two apps should be merged/colocated in some manner—maybe simply a jobs folder inside the api folder. Point being, they wouldn't work very well as "microservices" independent of each other so they should be coupled together.


apps/jobs/lib/data-cleanup/sync-leads.js
```diff
@@ -0,0 +1,73 @@
...
+  }
+})
+
+export const toClinikoUTC = time =>
+  moment(time)
+    .subtract(15, 'minutes')
+    .format()
+    .split('-')
+    .slice(0, 3)
+    .join('-') + 'Z'
```
craigspaeth: Seems like I need to figure out UTC format better, and why `moment().format()`'s UTC doesn't comply with Cliniko's b/c CI is returning something different from my local machine 🤔

lionsdwarf: Odd

lionsdwarf:
This all makes sense to me, @craigspaeth. Appreciate the concise write-up!

PR #312, opened 2018-11-05
[WIP] Cko scheduler killer

lionsdwarf:
@craigspaeth Here's the custom scheduler WIP. Some of these items may be redundant from slack, but a few things to note in your review:

1) Availability schema -- currently flat with a handful of fields on Availability, eg: Availability.bookingId. This is to say we are not currently utilizing fields on separate models and therefore the schema should/will be updated, eg: Availability.bookingId > Availability.Location.id. If this is wrong, pls advise.

2) In apps/api/models/availability.js, the "mergePractitionerAvailabilities" function is probably over-engineered, specifically with regard to the code block starting on line 112 that selects a practitioner when there is availability overlap. This code could be replaced with simpler but difficult/impossible-to-test random selection functionality. The current approach is verbose but guarantees a testable pure function. The extensive comments for this function will shed further light...

3) Availability.js currently following the "fromCliniko" function pattern. Not yet using "toCliniko" pattern, but will implement once schema revisions are implemented.

4) In our new react component, there is a "parseApptTypeId" function that will be replaced once we serve the apptTypeId from graphql, but right now we're cheating by parsing apptTypeId from calendarUrl.

5) Scheduler css - css import for 'react-datepicker' lib not working so there are no styles yet

That's all for now! Happy reviewing.

- JAL


apps/portal/components/schedule-appointment-calendar/index.js
```diff
@@ -0,0 +1,129 @@
...
+import DatePicker from 'react-datepicker'
+import 'react-datepicker/dist/react-datepicker.css'
+
+const NY_BIZ_ID = 69455
+
+const gql = new GraphQLClient(process.env.APP_URL + '/api')
+const parseApptTypeId = (url) => url.substr(url.lastIndexOf('=') + 1, url.length)
+const getApptTypeId = (props) => props.clinician ? parseApptTypeId(props.clinician.calendarUrl) : props.appointmentTypeId
+
+const getAvailabilities = async (props, date) => {
```
craigspaeth: 👍 I like how you split out the data-oriented here. This could even move to it's own `availability.js` "view model-like" lib alongside [the others](https://github.com/findOctave/cortex/tree/master/apps/portal/lib).

apps/portal/components/schedule-appointment-calendar/index.js
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  componentWillMount() {
+    this.setAvailabilities(this.state.userInputs.date)
+  }
+
+  render() {
+    return (
+      <div>
+        <DatePicker
```
craigspaeth: W00t! Excited to see how straightforward this is.

apps/portal/components/schedule-class/index.js
```diff
@@ -46,10 +47,13 @@ export default ({
         }
         {
           [
-            <ScheduleSchedule
-              content={props.schedule}
-              calendarUrl={props.calendarUrl}
-            />,
+            process.env.IS_CUSTOM_SCHEDULER ?
```
craigspaeth: 👌 nice!

craigspaeth:
I left a lot of comments, but it's mostly data-model/API/domain nitpicking—overall the code is looking really good! To address your questions in the PR description:

1. Generally speaking it's better to start with an object schema that you know is going to evolve even if it's as simple `availability { location { id } }` vs. `availability { bookingId }` because it's cheap + easy to continue to add fields to `location` vs. getting in a place where it's unclear and things get messy when people start adding `bookingName`, `bookingGeo: { long: String, lat: String }`. That said, no fields exposed is even better and I wonder if there's an opportunity to hide the concept of "booking" from our API as commented inline.
2. In this case, I do think we should go with the simpler/impure route (as I more extensively commented about inline)
3. 👍 I like it!
4. 👍 works for me, appreciate calling the temporary tradeoff out
5. No problem, happy to dig into CSS issues once we handoff the work to my plate. Appreciate putting in the `process.env.IS_CUSTOM_SCHEDULER` toggle to let us be iterative about tackling this.

👏 this is coming along nicely! My main hesitation is around #2 and making that `mergePractitionerAvailabilities` simpler. Feel free to address as much or little of the data-model nitpicking as you want. Good stuff!

apps/api/models/availability.js
```diff
@@ -0,0 +1,197 @@
...
+
+/*
+*  @param {2DArray} allPractitionerAvailabilities - array of arrays of Availabilities by Practitioner
+*
+*  The 2D array has to be flattened, ordered by start time, then made unique by start time
+*  The challenge is to perform the uniqueness operation and leave an even distribution of Practitioners associated with the "duplicate" start times
+*  The solution here uses a look-ahead stratgey, whereby duplicates are marked and practitioner ids are tracked in order to
+*  create, at minimum, a semblance of even Practitioner distribution
+*/
+export const mergePractitionerAvailabilities = allPractitionerAvailabilities => {
```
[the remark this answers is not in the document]
lionsdwarf: _.random will replace about 8 lines of code, but is probably still worth it. This enhancement, however, I'm de-prioritizing until post-Scheduler-launch, FYI

apps/api/models/availability.js
```diff
@@ -0,0 +1,197 @@
...
+        appointmentTypeId: consultApptType.id
+      })
+    )
+  )
+  return mergePractitionerAvailabilities(consultPractitionersAvailabilities)
+}
+
+export const availabilities = async (_root, args, { req }) => {
+  const isConsult = !Boolean(args.appointmentTypeId) && !Boolean(args.practitionerId)
+  const isClass = !Boolean(args.practitionerId) && Boolean(args.appointmentTypeId)
```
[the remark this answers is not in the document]
lionsdwarf: @craigspaeth On our next call, let's discuss this in light of the revised Availability schema.

PR #315, opened 2018-11-07
A couple charge job refactors

craigspaeth:
This does a couple small things related to data consistency I encountered in the charge job—I'll comment more inline.


apps/api/models/patient.js
```diff
@@ -10,7 +10,9 @@ const { REFERRAL_NOTE_TEMPLATE_ID } = process.env
 export const fromCliniko = patient => ({
   name: [patient.first_name, patient.last_name].join(' '),
   email: patient.email,
-  phone: patient.patient_phone_numbers[0].number
+  phone: patient.patient_phone_numbers.length
```
craigspaeth: Apparently, some patients don't have phone numbers in Cliniko. I think we assumed this was a requirement in our product but truth is we probably have folks being manually input into Cliniko and we probably shouldn't expect to have anything but an email—so this makes it a bit more defensive.

apps/jobs/lib/charge-sessions/network.js
```diff
@@ -34,56 +37,70 @@ export const addBillableDataToAppts = async appts => {
...
-        historicalCharges,
-        invoiceContent,
-        invoiceNotification,
-        lead,
-        patient,
-        practitioner,
-        stripeCustomer
-      }
-    })
+        if (!billableItem) return null
```
craigspaeth: I was often hitting Cliniko rate limiting so I slightly re-ordered the fetching so we can exit early if something required for charging is necessary (e.g. we don't fetch all the other things if the appt doesn't have a billable item).

apps/jobs/lib/charge-sessions/network.js
```diff
@@ -34,56 +37,70 @@ export const addBillableDataToAppts = async appts => {
...
+            customer: stripeCustomer.id
+          })).data
+          : null
+        const lead =
+          (await at.find({
+            table: 'leads',
+            filter: `emailAddress='${patient.email}'`
+          }))[0] ||
+          (await (async () => {
+            await syncPatients([patient])
```
craigspaeth: I still need to run the sync patients/lead job across all of our data, so until Airtable/Cliniko are in a consistent state, this code runs lazily here to ensure people who need to be charged get in a consistent state before doing so.

lionsdwarf:
Copy that! Thanks for the updates, Craig-o.

PR #319, opened 2018-11-09
Deploy

craigspaeth:


craigspaeth:
The code in master doesn't seem to reflect this bug, so not sure what happened, but going to deploy and stay alerted to this issue.

image

craigspaeth:
Turns out that was a red herring and the actual issue was simply some missing content from the prod. AT base.

PR #322, opened 2018-11-15
Limit parallel queries to accomodate Cliniko's rate limiting better

craigspaeth:
The hanging at the scheduler was due to the rate limiting plugin I added that was accommodating Cliniko's sadly aggressive rate limiting... we're probably going to have to eventually implement some kind of read store/caching for reading from Cliniko 😞


apps/api/models/appointment.js
```diff
@@ -182,9 +182,14 @@ export const appointments = async (_root, args, { req }) => {
     await Promise.all([
       cliniko.find({
         resource: 'individual_appointments',
-        sort: 'created_at:desc'
+        sort: 'created_at:desc',
+        limit: 10
```
craigspaeth: After returning > 100 appointment it would then query the attendees in parallel which would immediately hit the 60 reqs/minute rate limit.

PR #324, opened 2018-11-15
Sync leads continuously instead of chasing down data issues

craigspaeth:
This ties up some final loose ends with data sync issues like someone updating a lead's email in Cliniko and that getting out of sync with Airtable. It does this mainly by adding a new `upsert` function on the lead model which is used on create via our API and every 10 mins in the sync job. Everywhere else in the codebase where we we're querying for the Cliniko patient by email directly, I've replaced with first a query to Airtable, then a query by a new column in Airtable `clinikoId`.

I justify this process by using two concepts from database design:
1. Airtable is essentially our ["join table"](https://stackoverflow.com/questions/37016694/is-a-join-table-the-result-of-a-sql-join-or-the-table-between-a-many-to-many) that stores the joining IDs to Cliniko and eventually other relevant lead related data stores (e.g. Mailchimp). This uses Airtable as the source of truth for lead relationships and ensures we query by immutable/unique IDs instead of mutable data like email.
2. The sync-leads job acts as our ["database replication"](https://www.wikiwand.com/en/Multi-master_replication). It runs frequently/continuously to ensure the [multiple sources of truth](https://cdn.oreillystatic.com/en/assets/1/event/275/Managing%20multiple%20sources%20of%20truth%20in%20distributed%20applications%20Presentation.pdf) for shared lead fields (name, email, phone) are constantly queried, reduced to the current state of truth, and synced across systems.

More comments to come inline.


apps/jobs/lib/data-cleanup/sync-leads.js
```diff
@@ -52,24 +36,22 @@ export const fifteenMinsAgo = (now = moment()) =>
...
     })
   )
 }
 
 export default async () => {
-  const patients = await cliniko.find({
-    resource: 'patients',
-    q: `updated_at:>${fifteenMinsAgo()}`
-  })
+  const patients = await cliniko.find({ resource: 'patients', all: true })
```
craigspaeth: Updated this to go ahead an run across all patients because when this job synced out to Cliniko it would update the timestamp of the `updated_at` field on the patient—thus syncing them _again_ on the next go-around and forever-looping through the same patients 🙃

So with our small data-set I'm just going to punt on the "tailing the database logs" idea and run it over the entire "database".

apps/api/models/patient.js
```diff
@@ -27,12 +27,7 @@ export const toCliniko = ({ email, name, phone }) => ({
   ]
 })
 
-export const findOrCreate = ({ email, name, phone }) =>
```
craigspaeth: I'm dropping this `findOrCreate` functionality altogether for a unidirectional separation between reads and writes. 

We allow writes to occur via GQL mutations or the system's admin UI, and reads (GQL queries) will simply accept that the data is eventually consistent via the `data-cleanup` job instead of trying to lazily write data into a consistent state upon reading like `findOrCreate` does.

My hope is that this will avoid the spaghetti code of trying to bandaid inconsistent data wherever it happens to be read from, and centralizes the pipeline in one place. If we find that the eventually consistent data is still too out of sync (e.g. someone updates their email in their settings page and 5 seconds later it's not reflected in their Cliniko acct.), then we know we simply have to increase the frequency at which we poll for updates and sync to the "replicas" (in other words turn the Heroku Scheduler frequency up).

PR #329, opened 2018-11-19
Enhance avail model

lionsdwarf:
@craigspaeth Here is a good slice of the enhancements/cleanup to our Availability and Appointment schemas and queries. I went a little rogue renaming `...At` fields to `...Time` fields, eg `startAt` > `startTime`, as mentioned in one of my comment responses. If this isn't palatable, pls let me know and I can revert these.


apps/api/models/appointment.js
```diff
@@ -30,15 +36,15 @@ export const schema = {
...
-      createdAt: String
-      startAt: String
+      createdTime: String
+      startTime: String
       type: AppointmentType
       practitioner: Practitioner
       bookingIpAddress: String
       leads(bookingIpAddress: String): [Lead]
-      location: Location
+      studio: Studio
```
craigspaeth: We don't pay for Slack history right now, but I believe I polled for this and "location" was the term people preferred. I think "studio" is still debatable among the team because while it makes sense for a place that conducts classes, it's less ideal for describing the office or therapy clinic aspects. We also use "Our Locations" and such on the website copy.

If you did more domain work and I'm wrong feel free to keep this and update [our doc](https://docs.google.com/document/d/1l4uOgNuL_4Y56ESN9HhmY8PVMdzEm2z_d5khjsxKxjQ/edit)

apps/api/models/appointment.js
```diff
@@ -200,6 +209,39 @@ export const apptsBookedByIpAddress = (appts, attendeeGroups, ipAddress) => {
...
+  const patient = await cliniko.upsert({
+    resource: `patients`,
+    q: `email:=${args.leadEmail}`,
+    data: {
+      first_name: 'joe',
+      last_name: 'leinsdorf',
+      email: args.leadEmail
+    }
+  })
+  const appt = isClass
```
craigspaeth: ❤️ I like how this models closer to our domain: When you create an appointment at Octave you're either booking a slot in a class or booking a slot in a therapist/coach/intake coordinator's calendar. The underlying difference between attendees, individual/group appointments is an implementation detail with Cliniko. 👏  well done.

apps/api/models/availability.js
```diff
@@ -1,49 +1,37 @@
...
+      studioId: String!
     ): [Availability]
-  `,
-  mutations: `
-    bookAvailability(
-      appointmentStart: String
-      email: String
+    classAvailabilities(
+      date: String!
+      studioId: String!
```
craigspaeth: I think you can use the `ID` type in GraphQL here. This simply indicates that the string is unique and can be depended on by gql clients (e.g. Apollo client) for caching.

apps/api/models/appointment.js
```diff
@@ -97,10 +105,10 @@ export const toCliniko = apptData => ({
 
 export const fromCliniko = appt => ({
   id: appt.id,
-  endAt: appt.ends_at,
-  startAt: appt.starts_at,
+  endTime: appt.ends_at,
+  startTime: appt.starts_at,
```
craigspaeth: `At` to `Time` works for me! As long as we're consistent I'm not picky about the term.

craigspaeth:
Nice! This is a great domain modeling cleanup. My only nitpick was the change to "studio". If there's new evidence that's a better term to use then feel free to leave a comment about it and merge. 👏 well done!

apps/api/models/appointment.js
lionsdwarf: Ah, sorry, didn't realize that "location" had been vetted. I jumped the gun on this. Will revert to `location`. @craigspaeth

apps/portal/components/schedule/appointment-form/index.js
```diff
@@ -9,96 +9,152 @@ const MADISON_AVE_LOC_ID = process.env.MADISON_AVE_STUDIO_ID
...
+          }
         }
       }
-    `)
+    `
+  }
+}
+
+const getAvailabilities = async (props, date) => {
+  const formattedDate = date.format('YYYY-MM-DD')
```
craigspaeth: While Cliniko accepts this standard to do funny things like querying by wildcards e.g. `**-05-****` = all of may, we should be consistent about accepting consistent date/time formats across our API as UTC string.

apps/api/models/appointment.js
```diff
@@ -55,10 +61,12 @@ export const schema = {
   `,
   mutations: `
     createAppointment(
+      leadEmail: String!
       startTime: String
-      patientId: String
-      practitionerId: String
       appointmentTypeId: String
+      locationId: String
```
craigspaeth: For now, let's not expose this Cliniko implementation detail and hard-code the location ID to NY. Down the line we _may_ want to be explicit about the location of the appointment, but that will come with a lot more implications in the UI/UX so better to not cross that bridge too early.

apps/portal/components/test/schedule/confirm.test.js
```diff
@@ -0,0 +1,12 @@
+import { locationLabel } from '../../schedule/confirm'
+
+test('locationLabel converts a Cliniko business into a location', () => {
```
craigspaeth: 👍 Appreciate moving the test here! In the future, may I suggest trying to keep PRs smaller and more focused—ideally < 1k line diffs. This tripped me up in trying to merge your changes in because I didn't notice it amongst the the large diff/PR description. FWIW I'm totally guilty of this too—so please call me out too!

PR #330, opened 2018-11-19
Only query billable appointments up-front

craigspaeth:
Starting to hit Cliniko API rate limits left-and-right now that we're getting some non-trivial data in here. This helps the charge job run faster by filtering out appointments that aren't in an appointment type that has a billable item before fetching all the related individual pieces.


craigspaeth:
Going to start merging some of these because I want to get this code in prod. to run the first charging run.

PR #331, opened 2018-11-19
Continously sync created patients and airtable leads

craigspaeth:
This handles the null lead issue we saw today. I'm attempting to stick to my guns on the "separate read and write concerns + eventual consistency" plan here and avoiding upserting in the find by IP query.

This expands on the sync job to poll Cliniko for patients created in the last couple seconds (configurable). It will then sync over any new discoveries. I'm using `setInterval` instead of the Heroku Scheduler given the frequency of this job. This is closer to the real-time data syncing a database might do, so on that level, it's more ideal—however, given Cliniko's rate limiting it's unideal as it could clog up other requests the app is making.

That said, we probably won't need to do this once your in-house code takes over and one of my next tickets is to work around reads slowing down due to rate limiting. So we'll probably soon work around that throughput un-idealness.

I also adding syncing via Airtable to [address Murk's issue today](https://findoctave.slack.com/archives/CAASA9823/p1542650591016800).


craigspaeth:
I added this line to the comment as well:

> as well as querying newly created patients at a couple second intervals (for real-er-time needs in scheduling)

which is the case for the syncing of recently created patients

lionsdwarf:
Works for me! Thanks, @craigspaeth. Merge, baby!!

PR #332, opened 2018-11-20
Continously create snapshots of Cliniko resources for caching purpose

craigspaeth:
This gets the ball rolling on some caching practices we can use to work around Cliniko rate limiting. This will continuously (every 10 minutes alongside lead syncing for now) write snapshots of Cliniko endpoints to JSON data in PostgreSQL (or the no-install-overhead SQLite in dev).

We can then use these snapshots to return (at most 10 minutes old) stale data for read-only purposes. The hope is this continuous pipeline will avoid the invalidation problems that come with [traditional cache-aside](https://blog.cdemi.io/design-patterns-cache-aside-pattern/) patterns by changing it to an eventual consistency problem like we [deal](https://github.com/findOctave/cortex/pull/331) with [elsewhere](https://github.com/findOctave/cortex/pull/324).

I'm implementing this in querying appointment types for now as a test drive (because it's a simple query that gets actual traffic). If successful, down the line we can choose to query the database snapshots instead of Cliniko as we hit rate-limiting/performance problems.

More comments inline.


apps/api/models/appointment.js
```diff
@@ -181,7 +184,10 @@ export const findForLead = async lead => {
 }
 
 export const appointmentTypes = async _root => {
-  const apptTypes = await cliniko.find({ resource: 'appointment_types' })
+  const snapshots = await db('snapshot_cliniko_appointment_types')
+  const apptTypes = snapshots.map(
+    s => (_.isString(s.body) ? JSON.parse(s.body) : s.body)
```
craigspaeth: This is due to PostgreSQL and SQLite differences in storing structured JSON. As noted in the `db.js` comments we'll just require installing PG locally after we test drive this a bit b/c these dev/prod. discrepancies aren't worth maintaining in the long run.

apps/jobs/lib/data-cleanup/cache.js
```diff
@@ -0,0 +1,36 @@
...
+  }
+  log.success(`Cached Cliniko record ${doc.id} into ${tableName}`)
+}
+
+export const snapshot = async resource => {
+  const tableName = `snapshot_cliniko_${resource}`
+  const collection = await cliniko.find({ resource, all: true })
+  await db.schema.createTableIfNotExists(tableName, table => {
+    table.increments('id')
+    table.json('body')
```
craigspaeth: PG recently added first-class JSON support which is nice for being able to use it as a document store similar to MongoDB when needed. Since we're essentially just using PG as an ephemeral cache here I don't think it makes sense to do propper object-relational-mapping, and we can just try to do the most straightforward persisting of JSON blobs instead.

I do believe Octave will need to have more structured/relational data down the line, so I opted not to immediately reach for a NoSQL database if we can avoid having an extra DB in the ecosystem.

apps/api/models/appointment.js
lionsdwarf: Copy that!

apps/jobs/lib/data-cleanup/cache.js
lionsdwarf: Score!
SGTM, @craigspaeth

PR #347, opened 2018-11-26
Add UX elements to new scheduling flow

craigspaeth:
This puts things back in like disabling the next button, surfacing validation errors, and some new things like highlighting the clickable dates in the calendar.

cap


apps/portal/components/schedule/customer-form/index.js
```diff
@@ -76,26 +76,21 @@ const validateCouponForm = async (props, userInputs) => {
...
+            leadId: "${props.lead.id}"
+            stripeSource: "${token.id}"
+          ) {
+            name
           }
-        `
-      )
-      newState.isValidCard = true
-    } catch (err) {
-      throw new Error(err)
```
craigspaeth: Try to avoid `catching` where you don't need to. This line, for instance, is doing the exact same thing as if you didn't wrap it in a `try/catch` in the first place. I noticed in the lead-form you're [catching the error and returning it instead](https://github.com/findOctave/cortex/blob/master/apps/portal/components/schedule/lead-form/index.js#L52) too. This ends up swallowing the error and causes hard to trace bugs (e.g. if someone enters an invalid phone number it just silently fails).

PR #349, opened 2018-11-27
Fix h2s that need to hide

craigspaeth:
image


apps/portal/components/schedule/confirm/index.js
```diff
@@ -54,6 +54,8 @@ const Calendar = ({ confirm, appointment, p }) => {
 
 export default class Confirm extends Component {
   static type = 'confirm'
+  static isHideBodyHeaderH1 = true
```
craigspaeth: Followed the `isHideFooter` pattern here to hide the header content from accidentally rendering.

PR #355, opened 2018-12-03
Scheduler state management proof of concept using React hooks

craigspaeth:
Spent some time over the weekend trying out the new React hooks API to see what it would look like to do the whole "build up form state and send mutations at the end". Not intending to derail things and suggest we move fwd with this, but wanted to use it to spark conversation. More details to come.


apps/portal/components/schedule/appointment-form/index.js
```diff
@@ -1,284 +1,53 @@
...
-            id
-          }
-          practitioner {
-            id
-          }
-          location {
-            id
-          }
-        }
+export default () => {
```
craigspaeth: React Hooks allow us to use functional components everywhere. Which is a big win for separation of concerns, simplicity/consistency, and testability IMO.

apps/portal/components/schedule/view-model.js
```diff
@@ -0,0 +1,310 @@
...
+}
+
+// Borrowed from https://www.regextester.com/98583
+// TODO: If we do more complex validation we might want https://github.com/hapijs/joi
+const validatePhone = email => {
+  var re = /^(1\s?)?((\([0-9]{3}\))|[0-9]{3})[\s\-]?[\0-9]{3}[\s\-]?[0-9]{4}$/
+  return re.test(String(email).toLowerCase())
+}
+
+export const setLeadAttr = (state, name, val) => {
```
craigspaeth: With state management extracted it becomes a lot easier to write pure (Redux-like) functions of state -> new state.

apps/portal/components/schedule/view-model.js
```diff
@@ -0,0 +1,310 @@
...
+        ) {
+          name
+        }
+      }
+      `
+    )
+    // TODO: Apply coupon
+    return nextStep(state)
+  } catch (err) {
+    return { ...state, errorMsg: err.message }
```
craigspaeth: As you can see here, we've isolated all of the network code into this one flow where we can simply catch the error and pass it into state for rendering.

apps/portal/components/schedule/view-model.js
craigspaeth: I realized I forgot to commit my tests to this PR (from my machine at home), but these pure functions become very simple to write test coverage for. Simple state in-state out tests e.g.

````javascript
const inState = { ... }
const outState = setLeadAttr(inState, 'email', 'foo@bar.com')
expect(outState.lead.email).toEqual('foo@bar.com')
````

File: apps/portal/components/test/schedule/view-model.test.js [hunk 4]
```diff
@@ -0,0 +1,21 @@
...
+
+test('validateEmail passes for valid emails', () => {
+  expect(model.validateEmail('foo@bar.com')).toEqual(true)
+})
+
+test('validateEmail fails for invalid emails', () => {
+  expect(model.validateEmail('foo@a')).toEqual(false)
+})
+
+test('setLeadAttr stores a lead attr in state', () => {
```
craigspaeth: Here are those tests I was talking about.

PR #366, opened 2018-12-08
Referral page

mkandler:
Design updates to the referral page.

Still needs a small copy update: Could someone add the line
`Complete this referral form and we will follow up with your patient or client.`
as an `h2` on the contentModule `referral`?

(I don't have access anymore... ☹️)


craigspaeth:
Apologies @mkandler didn't realize you were still on the project and we started paying for Airtable so I was removing accounts to reduce 💰—I'll add you back. Thanks for this! Code LGTM!

mkandler:
Thanks @craigspaeth ! Not sure if you need to add me back on Airtable, I think this is the last little bit of the work we committed to do for launch. (I think...)

Hope all has been well with you!

PR #387, opened 2018-12-18
Snapshot Cliniko available times [WIP]

craigspaeth:
This implements continuously caching available times from practitioners. If we (most likely) hit limits with the performance of the new scheduler this should work around that by continuously saving snapshots of availabilities for the next week/year from the API to our database. This is in the _make it work_ stage, so once we hit performance issues I'll move it to the _make it right_ stage by adding tests/refactoring.


apps/jobs/index.js
```diff
@@ -28,6 +28,7 @@ schedule(etl.syncRecentlyCreatedPatients, JOB_INTERVAL_S)
 
 // Every couple mins
 schedule(etl.syncAssessments, JOB_INTERVAL_M)
+schedule(() => etl.snapshotAvailableTimes('week'), JOB_INTERVAL_M)
```
craigspaeth: The 7 day range limit is painful to work with and causes _tons_ of chatty throughput to the Cliniko API. Because of that we separate out quicker snapshots for the week with longer snapshots for the year (hoping that the week is the only one that needs to be more real-time).

craigspaeth:
Closing until we actually hit performance issues with this—which I imagine won't be too long from now 😭

PR #413, opened 2019-01-03
Fetch max number of charges from Stripe and other data scaling patches for charging job 

craigspaeth:
This fixes the horrible bug where the billing job tries to cross-reference past charges to avoid dupes but is limited to 10 so it doesn't find enough history to prevent all dupe charges.

The fix is simply to fetch as long a charge history as possible—which is 10k using the auto-paging API from Stripe's node interface. This should be more than enough because the charges are sorted by most recently created which means this job will query for the 10k most recent charges and cross-reference them with appointments for a patient in the last 90 days (and it's impossible that someone would attend over 10k appointments within 90 days).

I reviewed the data models involved for any other scenarios that could possibly cause similar data-scale related bugs and determined there are only a handful of "many" relationships in this web of appointments + group appointments + appointment types + practitioners + patients + customers + users + leads + billable items + charges + coupons + notifications. Those cases are...

1. Many appointments/group appointments
2. Many attendees in group appts.
3. Many coupons per attendee
4. Many charges per attendee

For which we should have our bases covered in this PR by:

1. Rate limiting and scoping by time frame should just mean the job gets slower with scale. Eventually, it may run out of memory and crash which will be okay b/c the actual "charging data mutation" happens at the very end. So crashing early is preferred (I would have loved for the job to crash instead of continuing to run after only fetching 10 charges). When that day does come we can scale up the machines or make the window smaller than the last 90 days of appts.
2. Cliniko enforces 100 max attendees, and we fetch their max page size of 100—so no adjustment needed here.
3. Couldn't find docs on the max number of relationships allowed in Airtable, but it's at least over 10 so I added a bit of code that crashes if a lead has >= 100 coupons just in case that doesn't play nice with page sizes elsewhere (highly unlikely anyway).
4. Fetching 10k of the most recent charges easily covers the last 90 days of appointments

So that should cover it for defensively coding for data scaling issues. Next up I will plan to put more safe-guarding "features" into the job to curb risk when it runs and limit the damage it can do in case something unforeseen does happen in the future.


lionsdwarf:
@craigspaeth Thanks for the detailed rundown. Stripe solution makes complete sense. Have a couple questions on the prophylactic measures for our next call.

PR #414, opened 2019-01-04
A couple big innocuous refactors

craigspaeth:
Sorry for the scary looking diff but this does a couple things that shouldn't visibly affect anything. This includes...

- Deleting the prelaunch app now that we're post-launch and don't surface any of those pages
- Partially applies the new scheduler code to the schedule session + standalone billing page + transfer client flows (mostly there but still need to handle some of the nuanced state logic—that's where those console.logs are for now)
- Replaces the global state module we were using with the more popular/reliable [reactn](https://github.com/CharlesStover/reactn)


craigspaeth:
Yeah, happy to tour you through the reactn / react hooks stuff—I'm pretty excited about the simplicity it. With these tools there's finally a lightweight/functional-friendly built-in state management solution similar to what [Baobab](https://github.com/Yomguithereal/baobab) provided long ago.

PR #415, opened 2019-01-04
Fully migrate new scheduler

craigspaeth:
This is an especially scary looking PR, but it will look much better after this one is merged https://github.com/findOctave/cortex/pull/414 This fully migrates the new scheduler code over and deletes the old scheduler code.


craigspaeth:
Going to clean this up a bit first

PR #417, opened 2019-01-05
Remove dead classes active and requiresConsult code

craigspaeth:
Just some small pieces of dead code cleanup.


craigspaeth:
This is just innocuous code deletion so mergin'

PR #419, opened 2019-01-05
Fix schedule session calendar

craigspaeth:
I accidentally hardcoded the time range when porting over to this flow. This fixes that by refactoring the scheduler state management a bit—it adds `onNextSessionMonth` to load practitioner availabilities (stored earlier by `onChoosePractitioner`). It does this when initializing and when clicking through months on the calendar.

Tangentially related I refactored the API that returns availabilities for a practitioner to add test coverage.


apps/onboarding/components/appointment-form/index.js
```diff
@@ -7,11 +7,10 @@ import { useScheduleState } from '../../lib/view-model'
 
 const cx = classNames.bind(styles)
 
-export default () => {
+export default ({ onNextMonth }) => {
```
craigspaeth: Since clicking through the months on the calendar behaves differently depending on the context (for consults it's across practitioners, scheduling a session is per-practitioner, and classes is actually based on class times) we move this to props instead of state so the different flows can configure the component differently.

PR #428, opened 2019-01-09
Fix deletes not syncing

craigspaeth:
Reported by Mego when she tried to delete an old class and it didn't reflect on the front-end (only two hard things right). It also refactors a bit to DRY things up and have better test assertions.


apps/jobs/lib/etl/cache.js
```diff
@@ -23,17 +23,18 @@ const upsert = async ({ doc, tableName }) => {
   log.success(`Cached Cliniko record ${doc.id} into ${tableName}`)
 }
 
-const createJsonTable = tableName =>
-  db.schema.createTableIfNotExists(tableName, table => {
+export const emptyAndUpsertAll = async (tableName, docs) => {
+  await db.schema.dropTableIfExists(tableName)
```
craigspaeth: Here we drop the snapshot table first, then create it again, before filling it up with snapshots. This _does_ cause a blip of missing data but I'm betting these transactions in PG with the little data we're dealing with doesn't make this a problem. We can look into making this operation more "atomically" sound down the line if it does prove to be a problem.

PR #435, opened 2019-01-09
Fix missing cookie bug

craigspaeth:
/profile on prod is throwing an error from `Cookie.parse` with an incognito window where there's no cookie existing. This defends against that.


craigspaeth:
This is only failing from the e2e test you're in the middle of fixing so gonna merge and deploy if that PR is 💚 .

PR #436, opened 2019-01-09
e2e test script, production-branch-specific e2e ci config

lionsdwarf:


.circleci/config.yml
```diff
@@ -1,5 +1,21 @@
 version: 2
 jobs:
+  build:
+    branches: 
+      only: 
+        - production
```
craigspaeth: Should this be only: master because it's a PR from master > production branches?

lionsdwarf: I think you're right @craigspaeth .

PR #441, opened 2019-01-10
Fix dupe coupon bug

craigspaeth:
Fixes the bug [Mego reported](https://findoctave.slack.com/archives/CEQMC5H1P/p1547078796004300)


craigspaeth:
Same e2e test ❌ , will make sure Deploy PR passes.

PR #446, opened 2019-01-10
Snapshot only the appointments that won't 404

craigspaeth:
After running `npm run job seed` against prod. data there were some 404s for practitioners that aren't marked `show_in_online_bookings` or for group appointments like classes. This adds some code that filters those out of being snapshotted.


craigspaeth:
Trying to get a deploy out, so mergin' but welcome any comments!

PR #450, opened 2019-01-10
Consultflowv3

lionsdwarf:
@craigspaeth Please see checkbox components in `apps/onboarding/components/services/index.js`


apps/onboarding/components/services/index.js
```diff
@@ -0,0 +1,57 @@
+import React from 'react'
+import styles from './index.css'
+import classNames from 'classnames/bind'
+import { xor } from 'lodash'
+import Checkbox from 'components/checkbox'
+import { useScheduleState } from '../../lib/view-model'
+
+const cx = classNames.bind(styles)
+
+const toggleService = (services, service) => xor(services, [service])
```
craigspaeth: Ooo `xor` I haven't used that before. I like the function here, but maybe it could be exported from view-model or even just branching logic inside `setLeadAttr` so it's easy to test.

apps/jobs/lib/seed.js
```diff
@@ -1,12 +1,12 @@
...
  * Jobs that need to run to seed some state initial state from a blank app/db.
- *
- * TODO: Consider this alongside database migration infrastructure
  */
 import * as etl from './etl'
 
-export default () =>
-  Promise.all([
-    etl.snapshotAvailableTimes('year'),
+export default async () => {
```
craigspaeth: Make sure you pull down the latest from upstream. This file should be deleted now—which might be related to the e2e running (I made a small change to refactor the `jest`/`jest-e2e` scripts so I could run `npm run jest filename` again, [see here](https://github.com/findOctave/cortex/blob/master/package.json#L17))

apps/onboarding/components/appointment-form/index.js
```diff
@@ -23,7 +24,12 @@ export default ({ onNextMonth }) => {
         {avails.map(availability => (
           <div
             key={availability.startTime}
-            onClick={() => selectAvailability(availability)}
+            onClick={() => {
+              selectAvailability(availability)
+              onAvailabilitySelect
```
craigspaeth: 👍 good composition here.

apps/onboarding/components/info-form/index.js
```diff
@@ -44,7 +44,7 @@ export default () => {
             onChange={event =>
               withinInfoForm('setLeadAttr', input.type, event.target.value)
             }
-            onBlur={() => allowLeadValidation(input.type)}
+            onBlur={() => setIsBlurred(input.type)}
```
craigspaeth: 👍 I like the clearer naming here.

apps/onboarding/lib/test/view-model.test.js
```diff
@@ -59,10 +110,19 @@ test('withinInfoForm ensure both lead and consent inputs are valid', () => {
       name: 'Karen H',
       phone: '555-555-5555'
     },
-    blurredLeadInputs: {
-      email: true,
-      name: true,
-      phone: true
+    leadValidations: {
+      email: {
+        isValid: true,
```
craigspaeth: I'm okay with this expanded state approach. However, as a general rule of thumb it's best to leave the minimum amount of state possible and use ["selectors"](https://github.com/reduxjs/reselect) or "derived state" because it can get easy to accidentally lose track of mutations when two pieces of state represent the same thing (similar problem to why it's encouraged to normalize your data in a database and have a single source of truth).

e.g. In this case, `isValid` is technically representing a value derived from `lead.email` and `blurredLeadInputs.email`. If you update `lead.email` and forget to update `isValid` at the same time you'll accidentally introduce a bug where the lead object's email is invalid but is represented and as valid.

This case seems pretty contained though, so I won't block the PR for it—just food for thought.

apps/onboarding/lib/view-model.js
```diff
@@ -83,31 +115,47 @@ export const withinInfoForm = (state, fnName, ...args) => {
...
-    return blurred && invalid ? 'Invalid phone number' : null
+export const leadValidationMsg = (state, fieldName) => {
+  const fieldValidation = state.leadValidations[fieldName]
+  return !fieldValidation ||
+    (!fieldValidation.isValid && fieldValidation.isBlurred)
+    ? FIELD_VALIDATION_MSGS[fieldName]
+    : null
+}
+
+export const isValidLeadInput = (fieldName, val) => {
```
craigspaeth: 👍 nice function composition going on here.

apps/onboarding/lib/view-model.js
```diff
@@ -300,6 +363,7 @@ export const submitConsultFlow = async state => {
     window.analytics.track('Consult Submitted')
     return nextStep({ ...state, appointment, lead })
   } catch (err) {
+    console.log(err)
```
craigspaeth: Nitpick: You can use `console.error` or usually I just like to leave this out of production code.

apps/onboarding/lib/view-model.js
```diff
@@ -468,13 +531,17 @@ export const submitting = state => ({
...
-export const allLeadInputsValid = state => {
-  const attrs = ['name', 'email', 'phone']
-  const isValids = attrs.map(attr => _.isNull(leadValidationMsg(state, attr)))
-  const isBlurreds = attrs.map(attr => !!state.blurredLeadInputs[attr])
-  const allValid = !_.includes(isValids, false)
-  const allBlurred = !_.includes(isBlurreds, false)
-  return allValid && allBlurred
+export const requiredLeadInputsValid = state => {
+  const requiredFields = Array.from(REQUIRED_LEAD_FIELDS)
+  for (let fieldName of requiredFields) {
```
craigspaeth: Nitpick: The functional programmer in me takes pause when I see `for` loops because, with loops, it's easy to lose track of values being mutated or counters are being stored in the loop. I find there's often a way to express something you want from a loop in a more concise/straightforward way using lodash functions or newer JS APIs. For instance, this whole function could be:

````javascript
_(state.leadValidations)
  .pick(REQUIRED_LEAD_FIELDS)
  .values()
  .map('isValid')
  .every()
````

Not a blocker either,  just food for though.

craigspaeth:
👏 Good stuff! I have some not blocking comments that you are welcome to refactor in a follow-up PR but since this code is clear and well tested I'm gonna merge. Well done!

apps/onboarding/lib/view-model.js
lionsdwarf: You're a lodash PRO @craigspaeth ! I hope to be so too one day.

PR #455, opened 2019-01-11
Drop the continuous snapshotting for cache-aside

craigspaeth:
Le sigh. After slogging through a couple bugs with my continuous snapshotting approach (inspired by [CQRS](https://martinfowler.com/bliki/CQRS.html)/[ES](https://martinfowler.com/eaaDev/EventSourcing.html) [projections](https://abdullin.com/post/event-sourcing-projections/) or [materialized-views](https://www.wikiwand.com/en/Materialized_view)). I've realized this is maybe not such a great pattern to try to out of the gate.

The truth is [cache-aside](https://vijayt.com/post/cache-aside-vs-read-through-cache/), while often riddled with issues (namely difficult application-managed cache invalidation), is probably the simplest pattern to implement. I don't think it makes sense to go the distance and try to implement a more robust pattern that requires things like seeding a DB, constructing a read data model, etc. until the cache invalidation issues of cache-aside become problematic.

So in this PR, I've updated our Engineering Philosophy section in the Readme to encourage an approach to data management that escalates in complexity choosing simple things like cache-aside then later continuous snapshots or w/e as needed—importantly, preferring nothing at all and querying a normalized single source of truth data store directly.

This then removes Knex and our DB dependency and simply adds a `cache: true` option to the Cliniko lib—which has a configurable [TTL](https://www.incapsula.com/cdn-guide/glossary/time-to-live-ttl.html) and uses [superagent cache](https://github.com/jpodwys/superagent-cache) to cache the specified reads in-memory (we can back it with Redis if our servers start to run out of memory).


craigspaeth:
Gonna merge and see how this performs. I welcome any comments!

PR #457, opened 2019-01-11
Only notify of a failed card once-per failed charge

craigspaeth:
Previously on every failed charge we would send an email prompting clients to update their CC. This would quickly get spammy because not only would it compound as multiple uncharged past appointments stacked up—it'll also run roughly daily, resulting in potentially dozens of repeated notifications a week.

To limit this spam we check to see if Stripe has a failed charge on record for this appointment already and if so we don't notify about it. This means we'll only ever send a failed charge notice once per-appointment.


lionsdwarf:
@craigspaeth Looks like it's checking Airtable, not Stripe, for the `failedChargeNotification` stamp, which makes more sense. 

Will merge.

PR #458, opened 2019-01-14
Consultflowv3

lionsdwarf:
@craigspaeth As suggested in your PR review, this enhancement utilizes "derived state" in lieu of setting the field validations directly on the state object.


apps/onboarding/lib/view-model.js
```diff
@@ -536,7 +517,7 @@ export const requiredLeadInputsValid = state => {
   for (let fieldName of requiredFields) {
     if (
       !state.leadValidations[fieldName] ||
-      !state.leadValidations[fieldName].isValid
+      !isValidLeadInput(fieldName, state.lead[fieldName])
```
craigspaeth: Neat! Didn't event realize you already had functions written that derived this state. Good stuff 👏

craigspaeth:
Good stuff! Thanks!

PR #461, opened 2019-01-14
Consultflowv3 qa

lionsdwarf:
This should have really been two separate PRs, but the changes contained within are as follows:
1) DOB formatting enforcement 
2) Schedule Appointment sub-header (configurable as H2 tag in Airtable markup)


craigspaeth:
Np grouping these together—they're both pretty small. LGTM, thanks!

PR #463, opened 2019-01-15
Fix input keyup issue on DOB

craigspaeth:
There's a `[type=date]` input that's probably more appropriate, but it's not supported in all modern browsers and iOS renders it weird. Since cleave is doing the heavy lifting, I'm going to keep things simple and just use this `pattern='/d*'` trick to limit to numbers https://stackoverflow.com/questions/25425181/iphone-ios-presenting-html-5-keyboard-for-postal-codes.


craigspaeth:
Actually one sec, found a better solution than tel.

PR #477, opened 2019-01-18
Refresh cko

lionsdwarf:
When nullifying Cko DOB and sex fields, only do so when record DOB === '2019-01-15'


craigspaeth:
👍 thanks!

PR #480, opened 2019-01-18
Fix timezone on billing receipt and other nice to haves

craigspaeth:
This does a couple small things to improve the billing job:
- Fixes the timezone on the receipt to be in ET time and say that in the description (had a client confused b/c their 8pm ET appt was recorded as the next day in GMT)
- Makes the "working hours" window dynamically based on the job interval (more inline)
- Splits out charging for old appts. months old from the more recent appts. (more inline)


apps/jobs/index.js
```diff
@@ -37,3 +37,4 @@ schedule(signupStageNotifications, JOB_INTERVAL_L)
 
 // A couple times a day
 schedule(chargeSessions, JOB_INTERVAL_XL)
+schedule(etl.backup, JOB_INTERVAL_XL)
```
craigspaeth: Forgot to add this to the worker in my last PR

apps/jobs/lib/charge-sessions/index.js
```diff
@@ -15,17 +15,22 @@ import moment from 'moment'
...
-    rollUpCharges(pastAppts).map(createCharge),
+    rollUpCharges(recentAppts).map(createCharge),
     toChargeWarnings(rollUpCharges(upcommingAppts)).map(sendChargeWarning),
-    rollUpInvoices(pastAppts).map(sendInvoice)
+    rollUpInvoices(recentAppts).map(sendInvoice)
   )
+
+  // Slowly run through old appointments to retry failed charges
+  const oldAppts = await getApptsWithBillables(threeMonthsAgo, twoWeeksAgo)
+  await rollUpCharges(oldAppts).map(createCharge)
```
craigspaeth: Running over historical appts. months ago is a really slow operation, so we're splitting this out from charging recent appts (within the last two weeks) to ensure the job runs fast and a deploy/restart to the worker doesn't interrupt it and cause a day's charges to potentially get skipped.

apps/jobs/lib/charge-sessions/validation.js
```diff
@@ -61,25 +61,29 @@ export const isInvoiceableAppt = (aggregate, now = moment()) => {
...
   const timezone = 'America/New_York'
-  const twelvePMEastern = momentTz(now)
+  const fivePMEastern = momentTz(now)
     .tz(timezone)
-    .hours(10)
+    .hours(17)
     .minutes(0)
     .seconds(0)
-  const fivePMEastern = momentTz(now)
+  const twoIntervalsBefore = momentTz(fivePMEastern)
```
craigspaeth: The intent for this window is not to make sure the job runs during work hours but that the actual charges occur before or during the work day so we can handle the complaints that morning/day. Basing the window on `JOB_INTERVAL_XL * 2`  ensures that the job still only runs once-per-day (if it runs every 6 hours then a 12 hour window ensures that the run following the first is out of the window), still preventing it from getting carried away.

PR #482, opened 2019-01-21
fix broken update cc query

lionsdwarf:
This resolves a bug introduced with the `auth` lib changes that prevented credit card updates from processing due to missing auth token.


lionsdwarf:
@craigspaeth Please see update containing tests

craigspaeth:
👌 awesome, thanks for the extensive tests!

PR #483, opened 2019-01-21
Apply styling to Auth0 lock widget

craigspaeth:
This adds styling to the Auth0 widget to make it more on-brand. It uses a mix of configuration we can send to the JS API and some unfortunate CSS overrides.

cap


apps/portal/pages/profile.js
```diff
@@ -12,10 +12,10 @@ export default class Body extends React.Component {
     return { user, email: query.email }
   }
 
-  async componentDidMount (props) {
-    this.setState({
-      user: this.props.user || (await getCurrentUser())
-    })
+  async componentDidMount () {
+    const user = this.props.user || (await getCurrentUser())
+    if (!user) return renderLogin()
```
craigspaeth: This is to ensure we open the login widget when landing on /profile directly (we point clients directly here when their CC fails so they can login and update it).

lionsdwarf: 👍

PR #484, opened 2019-01-22
Profile appts

lionsdwarf:
Displays an ordered list of upcoming appointments on a Patient's `/profile` page.

@craigspaeth With regard to querying upcoming appointments by start time, I did not wade into the timezone depths. Lets test and discuss. 

screen shot 2019-01-21 at 5 38 26 pm


apps/api/models/appointment.js
```diff
@@ -134,6 +136,22 @@ export const fromCliniko = appt => ({
...
+    resource: 'appointments',
+    q: [
+      `patient_id:=${patientId}`,
+      `appointment_start:>${moment()
+        .subtract(1, 'hour')
+        .utc()
+        .format()}`
+    ]
+  })
+  return appts
```
craigspaeth: This function looks great—good use of functional `sort` + `map` here! Nitpick: You could easily break out these 3 lines and the argument for `q` into pure functions with tests.

apps/api/models/client.js
```diff
@@ -67,7 +69,9 @@ export const fromAuth0 = user => ({
   creditCard: async () =>
     creditCardFromStripe(
       await stripe.customers.retrieve(user.user_metadata.customerId)
-    )
+    ),
+  appointments: async () =>
```
craigspaeth: Nitpick: Because you're immediately returning a promise, you don't need to have the `async` here (easy rule of thumb—if you don't see an `await` you don't need an `async`.

apps/api/test/models/appointment.test.js
```diff
@@ -102,3 +102,24 @@ test('addPatientToClass ignores added twice errors from Cliniko', async () => {
   })
   expect(appt).toMatchObject({ id: 'appt123' })
 })
+
+test('findUpcomingForPatient returns a list of upcoming appointments ordered by appt start time', async () => {
```
craigspaeth: 👍 appreciate the integration test.

Ideally, there'd be more of a "pyramid" of tests that cover more surface area (e.g. the two unit tests I describe above that specifically test sort/map logic and the `q` logic) but some integration tests can be a good backup plan in place of a full pyramid. My ultimate goal is always having _some_ coverage in PRs, so I won't block the PR for it.

apps/api/models/client.js
lionsdwarf: Good catch. Will resolve.

PR #485, opened 2019-01-22
Charge window is one interval wide

craigspaeth:
Mathz! 🤦‍♂️ Not sure why I thought making the window twice as big as the interval of the job would ensure the billing job ran once a day—that ensures it runs _twice_ a day! This fixes that incorrect mathz.

(FYI: The job always cross-references charges/appointments to avoid dupes so this isn't that problematic, but it just avoids spamming people with reminder emails multiple times a day when a charge fails).


craigspaeth:
I want to get this out before the next interval runs today, so gonna merge. As usual I welcome any comments!

PR #487, opened 2019-01-22
Shorten charging window to 48 hours

craigspaeth:
So we're more up to date with billing as requested per Product Planning earlier today.


craigspaeth:
Just a simple config change, going to roll out sooner than later to rip this bandaid off.

PR #492, opened 2019-01-24
Extract availability-picker into a sharable lib

lionsdwarf:


apps/api/models/appointment.js
```diff
@@ -66,6 +67,11 @@ export const schema = {
       practitionerId: ID
       appointmentId: ID
     ): Appointment
+    updateAppointment(
+      id: ID!
+      appointmentId: ID!
```
craigspaeth: This is redundant, it should just be `id: ID!` as we're updating `appointment` here.

apps/api/models/appointment.js
```diff
@@ -233,6 +239,25 @@ export const addPatientToClass = async ({ patientId, appointmentId }) => {
...
+const updateAppointment = async (_root, args, ctx) => {
+  authorizeClient(args, ctx)
+  let appt
+  try {
+    appt = await cliniko.update({
+      resource: 'appointments',
+      id: args.appointmentId,
+      data: {
+        starts_at: args.startTime,
+        // Null value for 'ends_at' will calculate the value according to ApptType duration
```
craigspaeth: 👍 thanks for the comment.

apps/onboarding/components/schedule-class.js
```diff
@@ -26,8 +26,11 @@ export default props => {
       header: <BodyHeader h1={props.content.appointment.h1} centered />,
       body: (
         <AppointmentForm
-          content={props.content}
-          onNextMonth={selectFirstMonthsAvailability}
+          onSelectAvailability={selectAvailability}
+          availabilities={state.availabilities}
+          selectedAvailability={state.selectedAvailability}
+          nextStep={nextStep}
```
craigspaeth: This seems like it shouldn't be a concern of this component. We should do this in `onSelectAvailability`, e.g.

````javascript
onSelectAvailability={() => {
  selectAvailability()
  nextStep()
}
````

File: apps/portal/components/appointments/index.js [hunk 4]
```diff
@@ -5,17 +5,22 @@ import moment from 'moment'
 
 const cx = classNames.bind(styles)
 
+const isWithin48Hrs = startTime => moment(startTime).isBefore(moment().add(2, 'days'))
```
craigspaeth: 👍 nice function! My instinct was this should be moved to "view model" or exported with a test but 🤷‍♂️ that wouldn't be testing more than `moment` is working—so not blocking for it.

lib/availability-picker/index.js
```diff
@@ -0,0 +1,101 @@
...
+  selectFirstMonthsAvailability
+} from './util'
+
+const cx = classNames.bind(styles)
+
+export default ({
+  onNextMonth,
+  nextStep,
+  state,
+  isProceedByFooter,
```
craigspaeth: Like `nextStep` this feels leaky and shouldn't be a part of this component. I think my comment above about consolidating in `onSelectAvailability` should remove the need for this too as you can decide outside the component if it should `nextStep` or not.

lib/availability-picker/util.js
```diff
@@ -0,0 +1,59 @@
...
+export const selectFirstMonthsAvailability = ({ availabilities, month }) => {
+  const [firstAvailability] = monthAvailabilities({ availabilities, month })
+  return firstAvailability
+    ? moment(firstAvailability.startTime).utc()
+    : moment(month)
+      .utc()
+      .startOf('month')
+}
+
+export const highlightedDaysClass = ({
```
craigspaeth: 👏 Nice work refactoring these functions! I think we lost a test for this one in the mix though.

lib/availability-picker/index.js
```diff
@@ -0,0 +1,101 @@
...
+  onSelectAvailability,
+  availabilities,
+  selectedAvailability
+}) => {
+  const [month, setMonth] = useState(moment())
+  const [selectedDate, setSelectedDate] = useState(
+    selectFirstMonthsAvailability({ availabilities })
+  )
+  const [loadingAvailabilities, setLoadingAvailabilities] = useState(false)
+  const [prevAvailabilities, setPrevAvailabilities] = useState(null)
```
craigspaeth: 👏 Good work with hooks here! It seems like there's a chance to consolidate these `useState`s into one blob e.g.
````javascript
useState({
  selectedDate: ...,
  loadingAvailabilities: ...,
  prevAvailabilities: ...,
  month: ...
})
````
...but if this isn't causing any issues then works for me!

**Comment by craigspaeth** [comment]
👏 nice work! Impressed you came out of the refactor hole so quickly. I'm mostly interested in [adding that test back in](https://github.com/findOctave/cortex/pull/492#discussion_r250778532) and [reducing the API surface area of `isProceedByFooter`/`nextStep`](https://github.com/findOctave/cortex/pull/492#discussion_r250776279).

File: apps/api/models/appointment.js (hunk 1 above)
Comment (lionsdwarf) [reply apps/api/models/appointment.js]: `id` is the User id, for Auth purposes

File: apps/onboarding/components/schedule-class.js (hunk 3 above)
Comment (lionsdwarf) [reply apps/onboarding/components/schedule-class.js]: Totally agree! Will refactor

File: apps/portal/components/appointments/index.js (hunk 4 above)
Comment (lionsdwarf) [reply apps/portal/components/appointments/index.js]: This is a WIP file. I agree. Will move it.

File: lib/availability-picker/index.js (hunk 5 above)
Comment (lionsdwarf) [reply lib/availability-picker/index.js]: Agreed!

File: lib/availability-picker/index.js (hunk 7 above)
Comment (lionsdwarf) [reply lib/availability-picker/index.js]: Got carried away with the new Hook syntax. Will reduce to a single `state` property.

File: lib/availability-picker/util.js (hunk 6 above)
Comment (lionsdwarf) [reply lib/availability-picker/util.js]: Not seeing a test for this in `master`. Will write one.

File: apps/api/models/appointment.js (hunk 1 above)
Comment (craigspaeth) [reply apps/api/models/appointment.js]: Ah I see, then we should use a different authorize function that checks `ctx.user.id` is equal to the appointment attendee's id—which will, unfortunately, result in a couple `cliniko.find`s before updating the appointment, but otherwise, we're not actually authorizing correctly here.

As it stands this will just check that the id passed in equal to the client ID but has no effect against whether they can update the appointment. e.g. anyone could pass in their client ID, pass in another person's appointmentId, and reschedule anyone else's appointment.

File: lib/availability-picker/util.js (hunk 6 above)
Comment (craigspaeth) [reply lib/availability-picker/util.js]: 😳 my bad then! I should have written a test for this 🙃

**Comment by craigspaeth** [comment]
Thanks for the replies! You actually pointed out another blocker about that authorization bit we should make sure to implement correctly.

File: apps/onboarding/components/schedule-class.js (hunk 3 above)
Comment (lionsdwarf) [reply apps/onboarding/components/schedule-class.js]: Gonna backtrack here on my agreement and explain why we need these two props, in addition to `onSelectAvailability`: 

`nextStep`: For Classes, there is no Footer to call this action. Instead it must fire from within the picker on Availability select
'selectedAvailability`: this is not an action, it's just the prop containing the Availability startTime.

Having said that, we could eliminate the `availabilities` and `selectedAvailability` props , which live on the `state` prop, so we are being a bit verbose/redundant, but I think we agreed that it makes sense to be explicit with the shape of data necessary to pass into the component.

File: lib/availability-picker/index.js (hunk 5 above)
Comment (lionsdwarf) [reply lib/availability-picker/index.js]: I'm removing the unnecessary `isProceedByFooter` prop and, instead, the component will call `nextStep` if the prop is provided, ie our footerless /Classes picker

File: apps/portal/components/appointments/index.js (hunk 4 above)
Comment (lionsdwarf) [reply apps/portal/components/appointments/index.js]: Before making this change, I'd like to discuss how best to slice the "view model" approach. I think we should avoid co-locating actions used by unrelated components within the same "view model" file, á la /onboarding.

File: lib/availability-picker/index.js (hunk 7 above)
Comment (lionsdwarf) [reply lib/availability-picker/index.js]: I've committed a broken refactor for this to a branch. I'm running into momentjs issues, and at this point today, after so many frustrating moments, I'm going to, for now, punt on this since it's mostly just code cleanup.

File: apps/api/models/appointment.js (hunk 1 above)
Comment (lionsdwarf) [reply apps/api/models/appointment.js]: The `id` is being passed to our `client.js` auth action. If this is not correct, let me know. 

https://github.com/findOctave/cortex/blob/e7e5e553bc0a78059de79d37147ef3ea6feb2a4c/apps/api/models/appointment.js#L243

File: lib/availability-picker/util.js (hunk 6 above)
Comment (lionsdwarf) [reply lib/availability-picker/util.js]: This function is just generating css classnames for the `monthAvailabilities`. I think our test for `monthAvailabilities` covers us here, but let me know if you disagree.

File: apps/onboarding/components/schedule-class.js (hunk 3 above)
Comment (lionsdwarf) [reply apps/onboarding/components/schedule-class.js]: Oh man, the late-day brain is weak. You're right, the nextStep call can live in onSelectAvailability. Revising now

File: apps/api/models/appointment.js (hunk 1 above)
Comment (lionsdwarf) [reply apps/api/models/appointment.js]: Ah, ok I follow you here, now. Will fix this. Thanks for highlighting.

File: lib/availability-picker/util.js (hunk 6 above)
Comment (craigspaeth) [reply lib/availability-picker/util.js]: Nope—good point! Now that you say it, I think that was my conclusion back then too 😛 Sorry for sending you in circles!

File: apps/portal/components/appointments/index.js (hunk 4 above)
Comment (lionsdwarf) [reply apps/portal/components/appointments/index.js]: Sounds good!

File: lib/availability-picker/index.js (hunk 7 above)
Comment (craigspaeth) [reply lib/availability-picker/index.js]: 👍 sounds good to me!

**Comment by craigspaeth** [comment]
👏 Good stuff! Thanks for the update. Lmk if you want to pair on the auth part.

File: apps/api/models/appointment.js [hunk 8]
```diff
@@ -233,6 +238,41 @@ export const addPatientToClass = async ({ patientId, appointmentId }) => {
...
+export const authorizeClient = async (patientId, apptId) => {
+  const appt = await cliniko.findOne({
+    resource: 'individual_appointments',
+    id: apptId
+  })
+  const appPatient = await cliniko.findOne({
+    resource: 'patients',
+    url: appt.patient.links.self
+  })
+  const user = await auth0.getUser({ id: patientId })
```
craigspaeth: This is correct code but `patientId` is confusing because you're actually passing the `ctx.user.id`—`userId` might be more obvious.

apps/api/models/appointment.js
```diff
@@ -233,6 +238,41 @@ export const addPatientToClass = async ({ patientId, appointmentId }) => {
...
+
+  if (user.user_metadata.patientId !== appPatient.id) {
+    throw new Error('Unauthorized')
+  }
+}
+
+const updateAppointment = async (_root, args, ctx) => {
+  await authorizeClient(ctx.user.id, args.appointmentId)
```
craigspaeth: 👍 good stuff.

apps/api/models/appointment.js
```diff
@@ -233,6 +238,41 @@ export const addPatientToClass = async ({ patientId, appointmentId }) => {
...
+      resource: 'appointments',
+      id: args.appointmentId,
+      data: {
+        starts_at: args.startTime,
+        // Null value for 'ends_at' will calculate the value according to ApptType duration
+        ends_at: null
+      }
+    })
+  } catch (err) {
+    throw new Error(err)
```
craigspaeth: Sorry, I didn't see this earlier, but this `try/catch` is unnecessary. Notice how you're throwing the exact same error you caught.`try` simply prevents an error within its braces from "throw"ing that procedure.

The only reason you would use try/catch is if you were going to do something different with the `err` than throw it right back (such as ignoring the error and logging it, transforming the error into a message appropriate to surface to the user, etc.). Most often this code lives "at the root" of your function calls (e.g. in our [jobs runner](https://github.com/findOctave/cortex/blob/master/apps/api/jobs/cli.js#L7)) or when something exceptional happens that we want to uniquely handle on the spot (such as [emailing people when a charge fails](https://github.com/findOctave/cortex/blob/master/apps/api/jobs/charge-sessions/network.js#L296)). When in doubt, just leave `try`/`catch` out, let the function throw, and let the error bubble up to one of these handlers.

PR #496, opened 2019-01-25
Schedule consult split test QA

craigspaeth:
Allows hiding coupon and better callout from the customer component. Small fix to clear error messages when going to the next step (noticed a stripe error would hang around on schedule page)

image


craigspaeth:
Simple config change. Going to merge to keep QA going.

PR #501, opened 2019-01-28
Reschedule appt

lionsdwarf:
@craigspaeth Contained within is reschedule functionality for a User's upcoming Consult and Therapy Appointments. I revised the dev roadmap we discussed together:

There was an issue requiring across Next apps. Accordingly, you will see a handful of copy/paste functions in `portal/components/reschedule-appointment/index.js` taken from the `/onboarding` view-model. 

Additionally, after building rescheduling functionality for Consults using copy/paste, I decided to do the same for the Therapy rescheduler. (This is opposed to building a full-blown `consult-picker` lib first (as previously outlined in our roadmap discussion)). I wanted to understand how much shared/unique functionality would exist across Consults <> Therapy before deciding where and how it makes sense to slice out these prop-management functions. 

This turned out to be a good approach bc, as you'll see in `portal/components/reschedule-appointment/index.js`, there are only two unique query functions stolen from the /onboarding view-model, plus two data formatting functions used in both query functions.

Let's discuss how best to DRY up these shared functions. One idea I'm leaning toward is a small `data-formatting` lib that can manage the inputs and outputs for the gql queries--I think it's probably overkill to separate out hard `consult-picker` and `therapy-picker` libs that extend our `availability-picker`. 

Good news is that both approaches are within reach without too much dev overhead, we just need to decide what makes most sense.

Please note that styles are still outstanding.


apps/portal/pages/profile.js
```diff
@@ -1,24 +1,87 @@
...
     const user = this.props.user || (await getCurrentUser())
     if (!user) return renderLogin()
-    this.setState({ user })
+    const appointments = await fetchUserAppointments(user.jwt)
+    this.setState({ user, appointments })
+  }
+
+  setView = async (viewData, refreshUserAppointments) => {
+    const state = {
+      ...viewData,
```
craigspaeth: The way this method generically takes and injects a `viewData` blob while being passed around to children seems off to me.

I'm not opposed to the general approach of storing `view: 'some-string'` in state to control views, but I wonder if it would be more straightforward to have this as global state, e.g. in [the "view model"](https://github.com/findOctave/cortex/blob/master/apps/onboarding/lib/view-model.js) so you don't have to pass this `setView` callback around. It might also be more straightforward to use specific functions per-view rather than attempting to accommodate each use-case with `setView`, e.g.

portal/components/appointments/index.js
````javascript
<a onClick={() => {
  await rescheduleAppointment()
  profileView()
}}>Reschedule</a>
````

onboarding/lib/view-model.js
````javascript
const profileView = async (state) => {
  const appointments = await fetchUserAppointments(state.user.jwt)) || state.appointments
  return { ...state, appointments, view: 'profile' }
}
````

That, or it might actually be more straightforward to have this be its own page that implicity reloads state with `getInitialProps`.

File: apps/portal/pages/profile.js [hunk 2]
```diff
@@ -1,24 +1,87 @@
...
+      // If rednering Profile componenet, ensure that the Upcoming Appointments list is fresh in case user just Rescheduled
+      appointments: (refreshUserAppointments && await fetchUserAppointments(this.state.user.jwt)) || this.state.appointments,
+    }
+    this.setState(() => (state))
```
craigspaeth: FYI: This can be `this.setState(state)`. At the least those extra parens around `(state)` are unnecessary and I think will be auto-corrected by prettier.

apps/portal/components/reschedule-appointment/index.js
```diff
@@ -0,0 +1,124 @@
...
+          toTime: "${to}"
+        ) {
+          startTime
+        }
+      }
+    }
+  `)
+  return { ...state, availabilities }
+}
+// END copy-n-paste
```
craigspaeth: Ah, interesting to see how just the loading of next month's availabilities end up being shared. I do think the data-formatting lib idea you have could make sense here. Looking forward to discussing more!

craigspaeth:
Looking good so far! I think my two larger takeaways are:

1. We should figure out how to best approach DRYing up that loading next month availabilities code—I like your lib suggestion
2. There's some state management we can try to consolidate between the component level vs. global "view model "level

Happy to pair through this soon! Thanks for sharing progress.

apps/portal/pages/profile.js
lionsdwarf: I'm with you on this--do not love the current `setView` function that receives an amorphous data blob. Considering that both Profile and Reschedule both need to load their own sets of distinct props, I'm feeling like the page approach will be the most intuitive / easy-to-manage. 

Will revise accordingly.

lionsdwarf:
@craigspaeth In the process of exposing a query for a single client appointment, I ran into an issue outlined in Slack. Accordingly, to avoid passing multiple appointment params via url, I'm fetching all client appointments and filtering client-side. Can revise as necessary . 

As discussed, our `auth` lib now contains an `authRequest` function for generating a gql request with jwt authentication headers. I added an additional `authClientQuery` function for generating a client query using the jwt. Leveraging these functions, the `updateAppointment` mutation now uses auth.

Styles are still outstanding. Please let me know if I should continue work on a query for single appointment.

apps/api/models/appointment.js
```diff
@@ -256,20 +256,15 @@ export const authorizeClient = async (userId, apptId) => {
...
-      data: {
-        starts_at: args.startTime,
-        // Null value for 'ends_at' will calculate the value according to ApptType duration
-        ends_at: null
-      }
-    })
-  } catch (err) {
-    throw new Error(err)
-  }
+  const appt = await cliniko.update({
```
craigspaeth: 👍 thanks for cleaning that up

apps/portal/components/appointments/index.js
```diff
@@ -12,16 +13,30 @@ export default ({ appts }) => {
...
     <div className={cx('appts')}>
       <h3 className={cx('header')}>Your upcoming appointments</h3>
-      {appts.map(({ startTime, type }) => (
-        <div className={cx('appt-card')}>
-          <div className={cx('day')}>{moment(startTime).format('DD')}</div>
-          <div className={cx('type')}>{type.name}</div>
-          <div className={cx('time')}>
-            {moment(startTime).format('dddd, MMM Do @ h:mma')}
+      {appts.map(({ startTime, type, id, practitioner }) => {
+        return (
```
craigspaeth: Nitpick: The `return` and curly braces are unnecessary here as you're immediately returning the JSX elements.

lib/auth/index.js
```diff
@@ -107,18 +107,35 @@ export const getJWT = req => {
...
+export const authGQL = jwt =>
+  new GraphQLClient(process.env.APP_URL + '/api', {
+    headers: {
+      authorization: `Bearer ${jwt.token}`
+    }
+  })
+
+export const authRequest = (query, req) => {
+  const { jwt } = getJWT(req)
+  if (!jwt) return null
```
craigspaeth: 👍 Nice! I like these functions. It might be better to `throw new Error('Not logged in")` here instead so it's more explicit than letting a `cannot access property foo of null` error throw later.

apps/portal/pages/reschedule.js
```diff
@@ -0,0 +1,31 @@
...
+    id
+    name
+    category
+  }
+}`
+
+export default class Reschedule extends React.Component {
+  static async getInitialProps ({ query, req }) {
+    const { client } = await authClientQuery(apptFields, req)
+    const appointment = client.appointments.filter(
```
craigspaeth: Lodash tip: You can use `_.find(client.appointments, { id: query.apptId })` to do this too.

apps/portal/components/reschedule-appointment/index.js
```diff
@@ -0,0 +1,68 @@
...
+      type.category === 'THERAPY'
+        ? await fetchSessionAvailabilities(
+          {
+            practitionerId: practitioner.id,
+            apptTypeId: type.id
+          },
+          date
+        )
+        : await fetchConsultAvailabilities({}, date)
+    setState(() => ({ ...state, availabilities }))
```
craigspaeth: Nitpick: This and [this line](https://github.com/findOctave/cortex/pull/501/files#diff-5bb97f40641de5fae64224df425bccf5R51) could be `setState({ ...state, availabilities })`

apps/portal/components/reschedule-appointment/index.js
```diff
@@ -0,0 +1,68 @@
...
+      onNextMonth={fetchAvailabilities}
+      onSelectAvailability={selectedAvailability =>
+        setState(() => ({ ...state, selectedAvailability }))
+      }
+      selectedAvailability={state.selectedAvailability}
+    />,
+    <button
+      disabled={!state.selectedAvailability}
+      onClick={async () => {
+        await rescheduleAppointment(id, state.selectedAvailability.startTime)
```
craigspaeth: 👍 good stuff! Nitpick: I wonder if these event handlers and `fetchAvailabilities` could be better extracted outside of this `default` function. It's a little tricky to read the mix of state management and JSX in this function. Not blocking by any means—just speculating.

apps/portal/pages/reschedule.js
```diff
@@ -0,0 +1,31 @@
...
+      appt => appt.id === query.apptId
+    )[0]
+    return { ...appointment }
+  }
+
+  render () {
+    return <RescheduleAppt {...this.props} />
```
craigspaeth: Happy to see how straightforward moving this to its own page is!

craigspaeth:
👏 Good stuff! My comments are all nitpicks/non-blocking/suggestions so merging! Thanks!

lib/auth/index.js
lionsdwarf: Thanks for handling this

apps/portal/components/appointments/index.js
lionsdwarf: Good call. Fixing now.

apps/portal/components/reschedule-appointment/index.js
lionsdwarf: Due to state update batching, the React docs encourage passing a function to `setState` instead of an object when the state update relies on the previous state. This is not the case here, so I've updated this code accordingly.

apps/portal/components/reschedule-appointment/index.js
lionsdwarf: Good call. About to submit revision to this that separates callouts from state updates.

PR #502, opened 2019-01-29
Added identify call to analytics so we can join GA events with AT leads

craigspaeth:
Marketing needs to answer the question "Did this lead join through consult split test A or B and through what source (e.g. direct traffic or FB ad)". The most straightforward way to do this would have been to add a column for `splitTestVariant` and `source` to Airtable and send them directly in the `createLead` mutation.

However, I believe this question is the tip of a larger iceberg around flexible "business intelligence" tools to be able to answer various questions like that in the future. Short of building a whole data warehousing stack I wanted to implement something minimal that allowed us to "join" data across services downstream instead of Engineering having to constantly play whack-a-mole sending the next piece of data Marketing needs to their service of choice.

[Segment](https://segment.com/docs/sources/website/analytics.js/) does a great job keeping a lot of this complexity out of the codebase and I eventually figured out that we simply need to send the piece of joining data to GA (in this case `leadId`) so that we can filter down a view that answers that question (data like `source` comes with all GA events).

So I did a bunch of work inside Segment/GA/Airtable control panels to result in this tiny change of code that sends `leadId` as a trait—or [custom GA dimension](https://segment.com/docs/destinations/google-analytics/#custom-dimensions)—in the [identify](https://segment.com/docs/spec/identify/) call. What's nice about the small resulting LOC and heavy config work is it's promising that non-engineers could do that majority of this work going forward.


lionsdwarf:
Nice work! Thanks for the overview and always nice to have a mostly-config solution.

PR #503, opened 2019-01-30
Ensure button is disabled until one selects an availability

craigspaeth:
This fixes a bug where an error would appear when someone tried to move past the scheduling step without picking an availability.


craigspaeth:
Gonna merge and get a deploy out but welcome your comments as always!

PR #505, opened 2019-01-30
Handle expired JWTs

craigspaeth:
This checks the `jwt.exp` property to see if the token is expired. If so, and it's on the client side, it'll call `logout`.


apps/portal/pages/profile.js
```diff
@@ -23,8 +23,9 @@ export default class Body extends React.Component {
 
   static async getInitialProps ({ query, req }) {
     const user = await getCurrentUser({ req })
-    const { client } = await authClientQuery(apptFields, req)
-    return { user, email: query.email, appointments: client.appointments }
+    const appointments =
+      user && (await authClientQuery(apptFields, req)).client.appointments
```
craigspaeth: There's an issue [in prod. at /profile](https://findoctave.com/profile) right now when one is logged out `Cannot read property 'client' of null` because we weren't checking if there was a user before attempting to make an authed query.

[Sure enough](https://github.com/findOctave/cortex/pull/501#discussion_r252339114) this was obscured because we were returning `null` instead of throwing a more explicit error. I went ahead and [implemented that here](https://github.com/findOctave/cortex/pull/505/files#diff-bce53ee2905fc531f965c44a5e09ad44R126).

lionsdwarf: Thank you, @craigspaeth . I saw that comment in my last PR but haven't gone through and implemented those changes yet. I just ran into this issue in my working branch. Thanks for fixing!

PR #506, opened 2019-01-31
Prevent 48hr resched

lionsdwarf:
@craigspaeth Prevents Appointment reschedule within 48 hours. Also contains a bug fix to prevent looped re-fetch of Availabilities when no Availabilities are returned.


apps/portal/components/reschedule-appointment/index.js
```diff
@@ -34,7 +37,7 @@ export default ({ startTime, id, type, practitioner }) => {
           date
         )
         : await fetchConsultAvailabilities({}, date)
-    setState(() => ({ ...state, availabilities }))
+    availabilities.length && setState({ ...state, availabilities: availabilities.filter(avail => !isWithin48Hrs(avail.startTime)) })
```
craigspaeth: Could we move this filter logic along with `isWithin48Hrs` to the util module and add a test? Othwerwise LGTM!

craigspaeth: An Enzyme test against the react component is welcome too—more concerned about a test than the refactor.

PR #507, opened 2019-01-31
Logged in mode QA

craigspaeth:
This does a ton of little random UI QA things. Some of the bigger changes involve:

- Extracting the marketing header into a shared nav component b/t portal + marketing (refactoring the props a bit to reduce surface area)
- Styling the rescheduler page
- Cleaning up some of the server/client rendering lifecycles states

There's a bit of copy paste occurring from taking the marketing header out, and I'm noticing some things are not ideal wrt. data fetching and how it interacts with the auth lib, but for now this gets the UI/UX in a decent place to launch logged in mode and we can look at cleaning that stuff up down the line.

screen shot 2019-01-31 at 1 45 16 pm
screen shot 2019-01-31 at 1 45 59 pm
screen shot 2019-01-31 at 1 46 13 pm


lionsdwarf:
@craigspaeth This is the PR I referred to during our pairing session. You left comments regarding not-ideal data fetching and auth lib interaction in our new Nav lib. Highlighting here as I've re-reviewed the PR and would love to discuss further on our next call.

craigspaeth:
Sounds good @lionsdwarf sorry that was a really vague comment. I was just thinking to myself that some of the data fetching parts of the auth lib are un-ideal because...
1. I discovered how convenient it would be to use the [syncronous version of getJWT](https://github.com/findOctave/cortex/pull/507/files#diff-af720d75792c8b84ac2708e244f05721R68) to determine if someone is logged in rather than have to fetch and inject that data down props/state
2. We might ideally want to [interoplate the query](https://github.com/findOctave/cortex/blob/master/lib/auth/index.js#L155) for current user data alongisde other initial queries to avoid roundtrips
These are minor things, but it made me pause and think there might be a better way. Happy to explain more on our next call.

PR #511, opened 2019-02-01
rm dupe function

lionsdwarf:
Removes a duplicate function and imports the function from our `availability-picker` lib


craigspaeth:
👍 good stuff, thanks!

PR #512, opened 2019-02-01
Book appt billing dep

lionsdwarf:
Booking a Class or Therapy requires successful customer creation / billing info submission.


apps/onboarding/lib/view-model.js
```diff
@@ -247,32 +252,39 @@ export const submitClassFlow = async state => {
...
-          endTime
-          startTime
-          location {
-            id
-            address1
-            city
-            state
-            zipCode
-            country
+    if (lead.id) {
```
craigspaeth: I don't think we want to wrap this in an `if`. If we fail to return a lead from `createLead` then we want to just let the function fail loudly instead of silently succeeding and not actually creating the client or appointment.

apps/onboarding/lib/view-model.js
```diff
@@ -247,32 +252,39 @@ export const submitClassFlow = async state => {
...
+      const { createClientFromLead: client } = await submitBilling({
+        ...state,
+        lead
+      })
+      if (client.email) {
```
craigspaeth: Similarly, since you're not using `client` further here I think it's okay to just an error bubble up from `submitBilling` to throw this function.

apps/onboarding/lib/view-model.js
```diff
@@ -247,32 +252,39 @@ export const submitClassFlow = async state => {
...
+        `
+        )).createAppointment
+        window.analytics.track('Class Submitted')
       }
-    `
-    )
-    await submitBilling({ ...state, lead })
-    window.analytics.track('Class Submitted')
+    }
     return nextStep({ ...state, appointment, lead })
```
craigspaeth: Given the `if` guards `appointment` here could be undefined if it doesn't succeed. Which, in the case of appointments, would fail to render the confirmation and Add to calendar button. This is a state we don't want to show in this case and would prefer to just let an error bubble up to the UI.

craigspaeth:
I think we simply want to swap the order of `await submitBilling` and the `createAppointment` mutation instead of putting these `if` checks along the way. That way if `submitBilling` fails it it'll throw right there and stop the rest of the function from executing.

The problem today is that the createAppointment mutation occurs first successfully then `submitBilling` throws—which means the UI correctly bubbles up the Stripe error but accidentally already booked the appt.

PR #513, opened 2019-02-01
Book Class and Therapy depend on successful customer creation

lionsdwarf:


craigspaeth:
LGTM! Ideally, we'd add a test here now that asserts the correct order of operations, but I know that test would be tricky and the ROI for it wouldn't make sense. Thanks!

PR #514, opened 2019-02-02
Logged in QA

craigspaeth:
More random logged in QA. The bigger pieces here are an optional progress bar in the nav to help with slow loading between pages, and truncating the availabiltiies list to avoid the picker from bouncing the page around.

image

cap

cap


lib/styles/index.css
```diff
@@ -1,19 +1,13 @@
 :root {
   /* Color */
-  --cliniko-bg-color: #575f5b;
   --coffee-color: #a6795b;
   --cream-color: #f9eecf;
   --dark-brown-color: #6f6150;
-  --dark-evergreen: #26391f;
```
craigspaeth: Someday I want to take a pass evolving the identity guidelines to give us more to work with for colors/typography—until then, less is more and these colors weren't being used in more than one place.

lionsdwarf: Copy that!

lib/components/nav/index.js
```diff
@@ -30,10 +31,17 @@ class Header extends Component {
...
+    // Track client-side page changes, close header if open on modal,
+    // toggle page-wide progress bar
+    Router.onRouteChangeStart = () => {
+      if (this.props.showProgressBar) this.setState({ progressBar: 'started' })
+    }
     Router.onRouteChangeComplete = () => {
       this.setState({ open: false })
+      if (this.props.showProgressBar) {
+        this.setState({ progressBar: 'finished' })
+        setTimeout(() => this.setState({ progressBar: 'done' }), 300)
```
lionsdwarf: What's the purpose of setTimeout here ?

PR #515, opened 2019-02-03
Support refresh ahead caching in Cliniko lib

craigspaeth:
We might want to hold on merging this until we discuss more but this implements a basic [refresh ahead cache](https://svenbayer.blog/2018/09/30/accelerate-microservices-with-refresh-ahead-caching/) option. It does this by replacing [the superagent cache plugin](https://github.com/jpodwys/superagent-cache-plugin) with a small home-grown implementation that caches the data in-memory and uses `setTimeout` to expire the cache based on TTL or `setInterval` to refresh the cache behind the scenes.

You can see in this screencap I'm using cache-aside with a TTL of 3 seconds via `cliniko.find({ cache: true })` which results in a slow response from a cache miss, followed by a fast response with a cache hit, then a slow response again from the cache being expired, and repeat.

a

In this screencap you can see refresh-ahead with a TTL of 3 seconds via `cliniko.find({ cache: { refreshAhead: true } })` results in a slow initial response from a cache miss, followed by only faster responses as the cache is being updated behind the scenes—resulting in no further cache misses.

cap

**Note:** Both screencaps are _not_ caching the `findOne` of the appointment type which is why the "fast" response is 300–500ms because it's still making that one roundtrip to Cliniko

This sounds like a nice silver bullet solution to get faster responses with up-to-date data, but let me put a warning out there that there's only [two hard problems in computer science](https://martinfowler.com/bliki/TwoHardThings.html) and this refresh-ahead magic could result in more edge case invalidation issues than the default cache-aside pattern.

So as with all things cache related, avoid at all costs!

Finally,

This is obviously missing tests and such, but I'm not sure this is a great idea yet and I'm tempted to try to roll this out to production and see how it handles real traffic before crossing the Ts and dotting Is. So apologies that this is not an example of stellar code quality practices!


apps/api/models/practitioner.js
```diff
@@ -46,7 +46,7 @@ export const fromCliniko = practitioner => ({
           practitionerId: practitioner.id,
           ...maxFromTo(page, fromTime, toTime)
         }),
-        cache: true
+        cache: { refreshAhead: true }
```
craigspaeth: Using refresh ahead on practitioner availabilities here, potentially speeding up rescheduling to an acceptable speed.

craigspaeth:
Going to give this a whirl and see how it performs. Made a ticket to delete or productionize it depending on how it performs. Thanks for looking over it!

PR #527, opened 2019-02-07
Salve satchel model

lionsdwarf:
The test code is way longer than the actual code. Love AT! Quick-dev for the win 🙌 

req_salvesatchel


lionsdwarf:
@craigspaeth Ordering functionality is in-progress, but I think this is still mergeable.

apps/api/models/salve-satchel.js
```diff
@@ -0,0 +1,38 @@
...
+  // For each salveSatchel, we make a request for the related salve records. This could be optimized
+  // by condensing all salve ids into a single AT request, then mapping the salves back to their
+  // related satchel.
+  return Promise.all(
+    salveSatchels.map(async salveSatchel => ({
+      ...salveSatchel,
+      salves: (await at.find({
+        table: 'salves',
+        id: salveSatchels.salves
+      })).map(airtableToModel)
```
craigspaeth: Nit pick: Might be more obvious to use a namespace here `import * as Salve from './salve'` and `Salve.airtableToModel`

craigspaeth:
💥 Easy peasy!

apps/api/models/salve-satchel.js
```diff
@@ -0,0 +1,38 @@
...
+  const salveSatchels = await at.find({
+    table: 'salveSatchels'
+  })
+  // For each salveSatchel, we make a request for the related salve records. This could be optimized
+  // by condensing all salve ids into a single AT request, then mapping the salves back to their
+  // related satchel.
+  return Promise.all(
+    salveSatchels.map(async salveSatchel => ({
+      ...salveSatchel,
+      salves: (await at.find({
```
craigspaeth: This could be a function to make this part a resolver so it doesn't have to upfront resolve salves if it's not requested in the GQL fields.

apps/api/models/salve-satchel.js
lionsdwarf: I like this

PR #528, opened 2019-02-07
find salve by id + order salves via AT UI

lionsdwarf:
Uses AT filter formula to request specific Salves by id.

Orders Salves according to related record order on Lead and SalveSatchel.

req_salvebyid

req_orderedsalves


apps/api/models/salve-satchel.js
```diff
@@ -23,13 +23,12 @@ export const salveSatchels = async () => {
...
-      ...salveSatchel,
-      salves: (await at.find({
-        table: 'salves',
-        id: salveSatchels.salves
-      })).map(airtableToModel)
-    }))
+    salveSatchels.map(async salveSatchel => {
+      return {
+        ...salveSatchel,
+        salves: await Salve.findById(salveSatchel.salves)
```
craigspaeth: 👍 nice. Just to note: You could replace the `await` here with a function `salves: () => Salve.findById...` to let it lazy-resolve. Not-blocking though b/c it's very unlikely you wouldn't request `salves` in a `satchel` 🎒

apps/api/models/salve.js
```diff
@@ -18,20 +19,32 @@ export const schema = {
...
+  image: args => airtableImgToResizeUrl(salve.image[0], args.width, args.height)
 })
 
+// If AT imposes a limit on filter characters, this filtered request
+// may need to be batched if we start requesting too many records
+export const findById = async salveIds =>
+  orderedSalves(
+    (await at.find({
+      table: 'salves',
+      filter: `OR(RECORD_ID()='${salveIds.join(`',RECORD_ID()='`)}')`
```
craigspaeth: Nice!

apps/api/models/salve.js
```diff
@@ -18,20 +19,32 @@ export const schema = {
...
+    })).map(airtableToModel),
+    salveIds
+  )
+
+// Salve ids on Lead are ordered (via AT UI), but "findById" does not return
+// return ordered results, so we manually re-order the Salve results according to the
+// id order on Lead.salves
+export const orderedSalves = (salves, orderedIds) =>
```
craigspaeth: 👍 good extraction of function here.

apps/api/test/models/salve.test.js
```diff
@@ -5,28 +5,72 @@ jest.mock('at', () => ({
...
+    id: 'yyy',
     title: 'Salve1',
+    label: 'Shamanistic herb ritual',
+    description: 'Not blank',
     type: 'audio',
     url: 'www.corn.com'
   })
 })
+
+test('orderedSalves returns salves in the related record order', () => {
```
craigspaeth: Solid tests!

craigspaeth:
👏 well done!

apps/api/models/salve-satchel.js
lionsdwarf: Missed this after going down the Ordering rabbit hole. Good call-will fix.

PR #529, opened 2019-02-07
salves resolver on salve-satchel

lionsdwarf:
@craigspaeth With the `salves` resolver, I had to cheat the test. Please let me know if I missed a better approach here:

https://github.com/findOctave/cortex/compare/master...lionsdwarf:salve-by-id?expand=1#diff-2c4923634e6c62e34563a5ada9923c9aR32


apps/api/test/models/salve-satchel.test.js
```diff
@@ -28,6 +28,8 @@ test('salveSachels resolves a list of salveSatchels with the related salve recor
     }
   ])
   const salves = await SalveSatchel.salveSatchels()
+  salves[0].salves = await salves[0].salves()
```
craigspaeth: This is fine. The important part is that you're calling the nested resolver like you're doing here `await salves[0].salves()`. The naming and injecting back into the response is a little counter-intuitive  though—it'd probably be better to set these to their own variables a and assert separately, e.g.

````javascript
const satchels = await SalveSatchel.salveSatchels()
const firstSatchelSalves = = await satchels[0].salves()
const secondSatchelSalves = = await satchels[1].salves()
assert(satchels).to(...)
assert(firstSatchelSalves).to(...)
assert(secondSatchelSalves).to(...)
````

But I'm less picky about tests being super clean/readable so not blocking. Thanks for the update!

lionsdwarf: While I was trying to mimic the real response, yours is a better approach--I agree.

PR #530, opened 2019-02-08
Display salves on Profile page

lionsdwarf:
In addition to Salve display, this PR contains a small modification to the auth lib to allow querying for non-client fields, and changes the `authClientQuery` signature for easier portability.

Styles are in progress...

screen shot 2019-02-08 at 4 31 42 pm


apps/portal/pages/profile.js
```diff
@@ -16,8 +15,48 @@ const apptFields = `appointments {
...
+    description
+    ${salveFields}
+  }
+`
+const contentFields = `
+  ${cmFields} 
+  ${satchelFields} 
+`
+const clientFields = `
+  ${apptFields}
```
craigspaeth: Nitpick: I think it'd actually be easier to read this if the query was in one big string (unless you're using the fragments elsewhere I'm not noticing).

apps/portal/components/profile/index.js
```diff
@@ -7,12 +8,13 @@ import styles from './index.css'
...
 const cx = classNames.bind(styles)
 
-export default ({ user, appointments, footer }) => {
+export default ({ user, appointments, footer, salve }) => {
   return (
     <div className={cx('profile')}>
       <Nav footer={footer} showProgressBar />
       <Settings user={user} />
       <Appointments appts={appointments} />
+      <Salves {...salve} />
```
craigspaeth: Mind putting a "feature flag" here `{process.env.IS_SALVE_BETA && <Salves {...salve} />}` so we can merge + deploys + staging QA rolling without accidentally shipping salves too early?

craigspaeth:
LGTM! Feel free to merge after adding that feature flag.

apps/portal/pages/profile.js
lionsdwarf: @craigspaeth I agree however, per your next comment, I will impose the "feature flag" at the query level, so we I'll leave this partitioned until we rm the flag.

apps/portal/components/salves-list/index.css
```diff
@@ -0,0 +1,74 @@
...
+      transform: translateY(-50%);
+    }
+  }
+  .text {
+    height: 133px;
+    background-color: #ffffff;
+    padding: 21px 24px;
+    border-radius: 0px 0px 8px 8px;
+    h5 {
+      font-family: Apercu;
```
craigspaeth: These type setting (font-size, family, weight, line-height) are encapsulated in mixins e.g. `@apply --apercu-caption` https://github.com/findOctave/cortex/blob/master/lib/styles/index.css#L53

craigspaeth:
LGTM! Don't forget about shared styles especially around typography (not a blocker). Thanks!

PR #532, opened 2019-02-11
handle server-side auth failure

lionsdwarf:


apps/portal/pages/profile.js
```diff
@@ -66,6 +67,13 @@ export default class Body extends React.Component {
 
   static async getInitialProps ({ query, req }) {
     const user = await getCurrentUser({ req })
+    if (!user) {
+      const gql = new GraphQLClient(process.env.APP_URL + '/api')
+      const footer = await gql.request(`query { 
+        ${cmFields}
+      }`)
+      return { footer }
```
craigspaeth: Fetching the footer shouldn't even be necessary here as we just [render a blank page when there's no user](https://github.com/findOctave/cortex/blob/master/apps/portal/pages/profile.js#L105),

apps/portal/pages/profile.js
```diff
@@ -99,7 +116,7 @@ export default class Body extends React.Component {
         user={this.state.user}
         appointments={this.state.appointments}
         footer={this.props.footer}
-        salve={this.props.salve}
+        salve={this.props.salve || this.state.salve}
```
craigspaeth: This probably shouldn't live in both state and props—but I can see why this is tricky with the whole server vs. client authed/unauthed tango going on here, so I'll defer this to the auth middleware refactor.

craigspaeth:
Works for me, thanks for the quick fix! We can clean this up later.

apps/portal/pages/profile.js
lionsdwarf: Removing this code breaks the app for reason's I've not yet dug into. We can investigate during cleanup.

PR #534, opened 2019-02-11
Salve est time

lionsdwarf:
screen shot 2019-02-11 at 11 42 59 am


apps/portal/components/salves-list/index.js
```diff
@@ -5,13 +5,35 @@ import SalveCarousel from '../salve-carousel'
 
 const cx = classNames.bind(styles)
 
-const SalveCard = ({ title, label, description, image, type, onClick }) => (
+const getVerb = type => {
+  switch (type) {
```
craigspaeth: 👍 nice

craigspaeth:
Super simple! Love how this is basically all config and no meaty logic—super quick to add new fields with this setp.

apps/api/models/salve.js
```diff
@@ -11,6 +11,7 @@ export const schema = {
       description: String
       type: String
       url: String
+      estimatedTime: String
```
craigspaeth: Nitpick: This might be more appropriate as a number representing milliseconds (e.g. in the case that we wanted to show a progress bar or some other numeric calculation). Since this whole effort is basically a prototype though I won't block for it.

PR #535, opened 2019-02-11
fix for lead without salves

lionsdwarf:


apps/api/test/models/salve.test.js
```diff
@@ -67,6 +67,11 @@ test('findById finds salves by id', async () => {
   })
 })
 
+test('findById returns an empty array if no salve ids are provided', async () => {
```
craigspaeth: Appreciate the test!

PR #542, opened 2019-02-12
Like Salve + Feedback

lionsdwarf:
"Like Salve" and "Feedback" UI + data funnel to Segment.

like-dislike-salve

feedback


apps/portal/components/salve-carousel/dislike.js
```diff
@@ -1,6 +1,6 @@
-export default () => (
+export default ({ fill }) => (
   <svg width='21' height='21' xmlns='http://www.w3.org/2000/svg'>
-    <g fill='#FFF' fill-rule='nonzero'>
+    <g fill={fill} fill-rule='nonzero'>
```
craigspaeth: You can also change this with css e.g. `svg { fill: red }`—and sorry to nit pick but I actually would prefer the background turn our "--yellow-color" instead of the fill of the icon (feels a bit Halloween-clunky to have equally balanced orange and black next to each other IMO).

apps/portal/components/salve-carousel/index.js
```diff
@@ -14,10 +14,65 @@ const Carousel =
     ? () => null
     : require('react-flickity-component')
 
+const Feedback = ({ collapseFeedback, salve }) => {
+  const [feedback, setFeedback] = useState()
+
+  const submitFeedback = () => {
+    window.analytics.track('salveFeedback', {
+      salve,
```
craigspaeth: 👍 component looks good, but I wonder if passing the whole salve object here will actually correctly send all that data to GA.

apps/portal/components/salve-carousel/index.js
```diff
@@ -14,10 +14,65 @@ const Carousel =
...
+      <div onClick={collapseFeedback}>X</div>
+      <div onClick={feedback && submitFeedback}>sbmt</div>
+    </div>
+  )
+}
+
+const LikeDislike = ({ salve }) => {
+  const [state, setState] = useState({})
+
+  const phil = isHighlight => (isHighlight ? '#e9982f' : 'white')
```
craigspaeth: 😂 I like the variable name `phil` here.

craigspaeth:
Mostly LGTM! Left a couple mostly nitpicks, and it'd be nice to get some kind of test going now that this component is getting more state/complexity—even the most basic Enzyme test would be good.

That said, this effort is kinda under "prototype" status so I'm less picky about the rigor—feel free to merge this test the things like the GA comment later if it's relevant.

apps/portal/components/salve-carousel/dislike.js
lionsdwarf: Just referred back to the rendering and, yup, I swapped the highlights. Will fix!

apps/portal/components/salve-carousel/index.js
lionsdwarf: Let's pair through that this pm.

lionsdwarf:
@craigspaeth Punting on the last couple css items. Thanks for taking the handoff here:
1) z-index issue (I made the sibling-component enhancement but there is still a small issue with the different page layers to iron out)
2) Arrow size on the 'No thanks' button

I fussed with the enzyme tests but ran into issues mocking the flickity lib. I am punting bc: POC and shallow/hooks issue.

craigspaeth:
Thanks for the cleanup!

PR #545, opened 2019-02-14
Salve time tracking

lionsdwarf:
Emit a tracking event every 15 seconds while user views a Salve. Sends `salveId` as event payload value. 

15sec


apps/portal/components/salve-carousel/index.js
```diff
@@ -92,6 +93,28 @@ export default class SalveCarousel extends React.Component {
...
+  componentWillUnmount () {
+    clearInterval(this.timer)
+  }
+
+  trackViewDuration () {
+    if (this.timer) clearInterval(this.timer)
+    if (this.props.index !== null) {
+      this.timer = setInterval(() => {
+        window.analytics.track(`salve${TRACK_INTERVAL / 1000}SecView`, {
+          value: this.props.salveCategory.salves[this.state.index].id
```
craigspaeth: This should be `salveId` instead of `value` so it's more obvious how to slice the tracking in GA.

apps/portal/components/salve-carousel/index.js
```diff
@@ -92,6 +93,28 @@ export default class SalveCarousel extends React.Component {
...
+    }
+  }
+
+  componentWillUnmount () {
+    clearInterval(this.timer)
+  }
+
+  trackViewDuration () {
+    if (this.timer) clearInterval(this.timer)
+    if (this.props.index !== null) {
```
craigspaeth: Surprised this guard needs to be here? Do we not pass `index` somewhere?

apps/portal/components/salve-carousel/index.js
```diff
@@ -92,6 +93,28 @@ export default class SalveCarousel extends React.Component {
     this.flkty.on('settle', () =>
       this.setState({ index: this.flkty.selectedIndex })
     )
+    this.trackViewDuration()
+  }
+
+  componentDidUpdate (prevProps, prevState) {
+    if (prevState.index !== this.state.index) {
```
craigspaeth: Nitpick: Instead of lifecycle hooks checking the state change I wonder if we should call `this.trackViewDuration()` inside `this.flkty.on('settle', () =>` instead. I think they achieve effectively the same thing though, so happy to have this—just sometimes hooking into lifecycles like this can get messy.

craigspaeth:
Mostly LGTM, the only blocker is the `salveId` bit. Feel free to merge after that.

apps/portal/components/salve-carousel/index.js
lionsdwarf: This needs to fire every time a user scrolls the carousel, so this must live as-is.

apps/portal/components/salve-carousel/index.js
lionsdwarf: You're correct. This is an artifact from a prior approach . Good eye!

PR #546, opened 2019-02-14
Jwt exp fix

lionsdwarf:
A propos our conversation today regarding "leaky" auth lib, thinking of moving data-fetching code into something like an `auth.fetch` export. What do you think about this approach ?

Thanks @craigspaeth !


lib/auth/index.js
```diff
@@ -75,12 +75,14 @@ export const extractTokenFromHash = href => {
   return params.id_token
 }
 
-export const logout = () => {
+export const logout = isClientSide => {
```
craigspaeth: This could be defaulted with `isClientSide = typeof window !=== 'undefined'`

lib/auth/test/auth.test.js
```diff
@@ -93,11 +93,41 @@ test('getJWT returns null when a jwt token a cookie is not provided in server re
...
+  const req = {
+    headers: {
+      cookie: 'str'
+    }
+  }
+  JWTDecode.mockReturnValueOnce({ sub: 'foo', exp })
+  JSCookie.get.mockReturnValueOnce('authtoken=foobar')
+  const jwt = auth.getJWT(req, null, moment.unix(exp + 1000))
+  expect(jwt).toBeFalsy()
+  expect(JSCookie.remove).toBeCalled()
```
craigspaeth: I see the issue here is basically an expired token server-side—throwing an error before it can be logged-out client-side? It's odd this works then, because JSCookie is a client-side lib. I wonder if JSCookie is just not running server-side, and something else it allowing the error not to surface. To clear a cookie server-side we'd have to call [something like `res.clearCookie()`](https://stackoverflow.com/questions/27978868/destroy-cookie-nodejs) on the server.

craigspaeth:
Happy to have this fix, but I wonder if we'll have to circle back on this library to clean up the way it deals with isomorphic stuff.

lib/auth/test/auth.test.js
lionsdwarf: Oh weird. Good point.

lionsdwarf:
Let's see what shakes out. Thanks Craig.

PR #547, opened 2019-02-15
[WIP] Profile tabs

lionsdwarf:
@craigspaeth Do you know of a better way to do this aside from the MediaQuery component ?


craigspaeth:
👍 no the MediaQuery component is great! I would have basically suggested implementing that component manually by setting screen size in state or hiding/showing elements based on CSS media queries.

PR #548, opened 2019-02-15
Profile tabs

lionsdwarf:
Profile tabs for mobile. 

tabs_portrait


apps/portal/components/profile/index.css
```diff
@@ -29,5 +29,31 @@ main {
 }
 
 .container {
-  margin: auto;
+  @media (min-width: 768px) {
```
lionsdwarf: @craigspaeth Not sure why but when I used `@media (--tablet-portrait)` here it rendered as max-width instead of min-width in the browser, even though this is a min-width variable

craigspaeth: 🤷‍♂️ that's weird! I'm fine with the hardcoded breakpoint—IMO, breakpoints aren't something to get too DRY-happy with. Sometimes the content warrants breaking at a very specific point where text starts to awkwardly overlap/rag vs. some predefined screen-size. So 👍 LGTM!

lionsdwarf: 👍

PR #550, opened 2019-02-15
revise value prop in tracking calls

lionsdwarf:
@craigspaeth `feedback` still not coming though to GA but maybe the new dimension config hasn't yet propagated to Segment's CDN ? Idk


craigspaeth:
LGTM! Did you make sure to add the new dimension number in Segment?

lionsdwarf:
Yessir we are updated in Prod !

PR #554, opened 2019-02-18
Carousel swipe polish

craigspaeth:
Ah, fun times trying to get a web app to have native-like UI patterns. There're a couple of challenges to get this iframe embedding to feel natural on mobile:
1. Momentum when vertically scrolling
2. Interact with elements inside iframe, but swiping right/left at the edges is still possible
3. iOS swipe left inside a browser will by default go to the previous webpage—we made the carousel "wrap around" so one can just swipe left to get back to the beginning

For 1 I used `-webkit-overflow-scrolling: touch;` in a previous PR, for 2 I added these left and right "hit" areas (pictured below) that when tapped-down grow to 100% width to allow Flickity to do it's job, for 3 I simply added `wrapAround: true` to Flickity.

cap


apps/portal/components/salve-carousel/index.js
```diff
@@ -94,7 +94,7 @@ export default class SalveCarousel extends React.Component {
   }
 
   componentDidMount () {
-    this.flkty.on('settle', () =>
+    this.flkty.on('change', () =>
```
craigspaeth: It feels more natural to have the number change on this event as opposed to when momentum ends.

lionsdwarf: 👍

PR #555, opened 2019-02-18
Add to homescreen

lionsdwarf:
Foregoing tests on this functional component that uses hooks

add_to_home


craigspaeth:
LGTM! I intended for the button to persistently live above the tabs, but this should be easy enough to move once I get in there and design the actual modal. Thanks!

PR #561, opened 2019-02-19
sync clients

lionsdwarf:


craigspaeth:
Mostly LGTM! It would be great to get some logging in there—at least when the job runs and finishes, but maybe also after each client update. Feel free to merge when that's in and when the build it 💚

apps/api/jobs/etl/sync-clients.js
```diff
@@ -0,0 +1,52 @@
...
+})
+
+export const groupPersonData = ({ patients, customers, users }) =>
+  users.map(user => ({
+    user,
+    patient: _.find(patients, { id: user.user_metadata.patientId }),
+    customer: _.find(customers, { id: user.user_metadata.customerId })
+  }))
+
+export const filterSyncGroups = groups =>
```
craigspaeth: 👍 nice thanks for taking it the extra step!

craigspaeth:
👏 good stuff! Thanks for the updates.

PR #563, opened 2019-02-20
write Pronouns to gender_identity instead of sex field

lionsdwarf:


craigspaeth:
Bloody good on ya Cliniko mates!

PR #572, opened 2019-02-28
Auth request

lionsdwarf:
Builds atop the `carouselless` branch.

Moves authenticated request logic into `auth` lib, which greatly reduces the need for boilterplate as well as multiple dependency imports. Authenticated requests can now be made with 2 lines of code, total. 

If we feel this makes `auth` lib leaky, we can divide into `server` (middleware) and `client` (data-fetch, Lock) sub-libs, as previously proposed.


apps/portal/pages/profile.js
```diff
@@ -75,22 +73,9 @@ const getQuery = jwt => `
...
-    const jwt = JWTDecode(auth0Token)
-    const gqlQuery = getQuery(jwt)
-    const options = {
-      headers: {
-        authorization: `Bearer ${auth0Token}`
-      }
-    }
-    const gql = new GraphQLClient(process.env.APP_URL + '/api', options)
-    const { client, footer, salveCategories } = await gql.request(gqlQuery)
+    const { gql, jwt } = authRequest(req)
```
craigspaeth: 👍 I like it This feel like the right level of abstraction—I think it was leaky before because we were trying to encapsulate parts of the GQL query and this wraps minimal boilerplate.

PR #589, opened 2019-03-04
Send notifications upon assigning salve

craigspaeth:
This implements notifications upon assigning a salve by following basic patterns from our signup stage notifications (a `salve.sentNotificationsTo` column in AT keeps track of who received notifications). It also comes with a couple of small refactors:

- We can explicitly extend notification variables with a `vars` argument (e.g. mustache variables like `See content here: {{salveUrl}}`)
- Wrapped signup stage + assigned notifications in index file
- Extracted `sendEmail` and `sendSMS` into their own "network" module


apps/api/jobs/notifications/assigned-salve.js
```diff
@@ -0,0 +1,57 @@
...
+import * as log from '../log'
+
+const { APP_URL } = process.env
+
+export const toGroups = salves =>
+  salves
+    // Find salves with unsent notifications
+    .filter(
+      salve =>
+        !_.isEqual(_.sortBy(salve.leads), _.sortBy(salve.sentNotificationsTo))
```
lionsdwarf: Nice use of `sortBy` for the comparison !

apps/api/jobs/notifications/assigned-salve.js
```diff
@@ -0,0 +1,57 @@
...
+  const sendingNotifications = groups.map(
+    async ({ salveUpdate, leadIds, salveUrl }) => {
+      const leads = await Promise.all(
+        leadIds.map(id => at.findOne({ table: 'leads', id }))
+      )
+      await Promise.all(
+        leads.map(async lead => {
+          await sendEmail(lead, notification, { salveUrl })
+          await sendSMS(lead, notification, { salveUrl })
+          await at.update(salveUpdate)
```
lionsdwarf: Here do you think we should await send success before updating AT ? I mention this in light of the synchronicity issues we've observed in our create user & client calls, but I realize the data integrity here for notifications is low risk...

PR #598, opened 2019-03-07
Pass custom vars through SMS

craigspaeth:
I notice SMS wasn't allowing the `{{salveUrl}}` injection in copy like email was. This fixes that.


lionsdwarf:
I think you need a new mock for AT to make the build 💚 . Then merge-away !

PR #618, opened 2019-03-08
Support appt cancellation

craigspaeth:
This was actually a pretty simple feature to build despite the noise in this PR. What's going on here is...

- Delete appt mutation added to API (super simple given you already wrote + tested `authorizeClient`)
- Refactored Appointments component to live inside Settings and moved `appointments` to state
- Added view model function that send mutation and state update
- Styled up some UI to make it look good

cap


apps/portal/components/appointments/index.js
```diff
@@ -1,15 +1,20 @@
...
+import Trash from './icon-trash'
 
 const cx = classNames.bind(styles)
 
 const isWithin48Hrs = startTime =>
   moment(startTime).isBefore(moment().add(2, 'days'))
 
-export default ({ appts }) => {
+export default () => {
+  const { state, cancelAppointment } = useSettingsState()
```
lionsdwarf: In some cases I've started treating lines of code like this one--where element-specific state and methods are defined--as indicators that a separate sub-component might be appropriate instead. It's superfluous in this simple component with not a lot happening inside, but when these component trees start to grow and a single state instance starts to manage a bunch of unrelated element states, it gets a lot easier to reason about the code when there are multiple little sub-components, each of which encapsulates it's own state. This approach can also greatly reduce the complexity of derived state calculations, or eliminate them entirely.

Just food for thought, and definitely not important here. Thanks for reading my rant.

lionsdwarf:
LGTM! Glad this was easy peasy. Merging....

PR #649, opened 2019-03-12
[Don't merge WIP] Sync G Sheets assessments instead of AT

craigspaeth:
Phew. Thanks for directing to me the `googleapis` module—that was way easier to set up than slogging through permissions trying to figure out Google API's cpanel for [google-spreadsheets](https://www.npmjs.com/package/google-spreadsheet). This changes the assessment sync job to convert a sample google sheet Murk gave me into the same `assmt` looking object we got from Airtable—keeping a lot of the code the same.

Still todo:
- [x] Rename `table` to `assmtTypeName`
- [x] Refactor + tests
- [x] Ensure it doesn't double sync assessments


apps/api/jobs/etl/sync-assessments.js
```diff
@@ -13,12 +13,34 @@ import _ from 'lodash'
...
+auth.setCredentials(JSON.parse(process.env.GOOGLE_SHEETS_TOKEN_JSON))
+const sheets = google.sheets({ version: 'v4', auth })
 
-// the note id exists only at the end of the url link
-export const parseClinikoNoteIdFromLink = link =>
-  link.substring(link.lastIndexOf('/') + 1, link.length)
+const fetchSheetData = spreadsheetId =>
+  new Promise((resolve, reject) => {
+    sheets.spreadsheets.values.get(
+      { spreadsheetId, range: 'A1:999999' },
```
lionsdwarf: What's this range property for ?

craigspaeth: It's like the formula for selecting a range of data in the spreadsheet. This selects column 1 row A through ~1mil—aka the entire data set I figure by the time we have more data than selectable with formulas/in-memory we'll be building this in-house.

craigspaeth:
Looks like tests passed but not the staging deploy. I think this is still good to merge then.

lionsdwarf:
That shouldn't happen. I made a change to the yml file yesterday that I think caused this. I'll fix--thanks for the heads up.

PR #677, opened 2019-03-14
aptible ssh instructions

lionsdwarf:
@craigspaeth Not sure this should live here, but if not, let's discuss where we should put this kind of info.


craigspaeth:
README is good by me! If it gets too much we can consider a /docs folder, wiki, or just leaving it in the Notes in the Product Trello.

PR #679, opened 2019-03-15
[WIP] Sync lead.id, lead.receivedNotifications, lead.receivedSalveNotifications, lead.salves

lionsdwarf:
A `sync-person` job that syncs AT Leads to psql as follows:
1) Lead 'id' > `lead_id` column on `person` table
2) Lead `receivedNotifications` > `person_onboarding_notification` table
3) Lead `receivedSalveNotifications` > `person_salve_notification` table
4) Leads 'salves' > `salve_id` on `person_salve` table

Contrary to our initial roadmap, I avoided modifying `sync-clients` and instead created a separate job for this sync--I feel like we should opt for smaller, specialized sync jobs rather than convoluting existing sync jobs with new psql logic. 

Leverages a "bulk insert if not exists" method for psql--I figured it was worth generating the dynamic raw SQL generator function contained within as it will be used across sync jobs going forward (I was unable to find a lib to generate this kind of raw query).

Note: this should only be merged AFTER the "pg schema" PR is merged.


apps/api/jobs/etl/sync-person.js
```diff
@@ -0,0 +1,67 @@
...
+const syncPersonSalveNotification = (people, leads) => pg.bulkInsertIfNotExists({
+  table: 'person_salve_notification',
+  unique: ['person_id', 'salve_id'],
+  records: personIdToReceivedNotificationIds(people, leadIdToNotifiedSalveIds(leads), 'salve_id')
+})
+
+const syncPersonOnboardingNotification = async (people, leads) => pg.bulkInsertIfNotExists({
+  table: 'person_onboarding_notification',
+  unique: ['person_id', 'notification_id'],
+  records: personIdToReceivedNotificationIds(people, leadIdToNotificationIds(leads, await at.find({table: 'notifications'})), 'notification_id')
```
craigspaeth: Ditto note about the intermediate maps here with `leadIdToNotificationIds`.

apps/api/jobs/etl/sync-person.js
```diff
@@ -0,0 +1,67 @@
+import * as pg from 'pg-client'
```
craigspaeth: Nit pick: Because there's already a module in our package.json called `pg` could we name this something else like `import * as db from 'db'` or `import * as db from 'db'`. It was slightly confusing to see `pg.bulkInsertIfNotExists` and find out that wasn't a method on the driver itself.

apps/api/jobs/etl/sync-person.js
```diff
@@ -0,0 +1,67 @@
...
+})
+
+export default async () => {
+  const leads = await at.find({ table: 'leads' })
+  await syncPerson(leads)
+  const people = await pg.knex('person')
+  await syncPersonSalveNotification(people, leads)
+  await syncPersonOnboardingNotification(people, leads)
+  await syncPersonSalve(people, leads)
+  process.exit(0)
```
craigspaeth: This shuts down the whole node process (e.g. it'll kill the worker once this function runs). I assume you used this to shut down this script locally b/c it keeps a connection to PG open. Maybe we can wrap a `db.closeIfDev()` function or something, but regardless this line should be removed to keep the worker running in production.

apps/api/test/jobs/etl/sync-person.test.js
```diff
@@ -0,0 +1,175 @@
+import * as pg from '../../../jobs/etl/sync-person'
+
+jest.mock('at', () => ({
+  find: jest.fn()
+}))
+
+describe('toPGPerson', () => {
```
craigspaeth: 👌 great tests!

lib/pg-client/index.js
```diff
@@ -0,0 +1,45 @@
...
+    : `1 FROM ${table} t WHERE t.${unique} = d.${unique}`
+
+/*
+*  Build raw SQL string to insert a collection of records for those that do not already exist in table, aka "bulk insert if not exists"
+*
+*  @param {String} table = psql table name
+*  @param {Array<Object>} records = list of object records to be inserted into psql with properties matching psql columns names
+*  @param {String || Array<String>} unique = string or array of strings matching the column or columns on which the sql statement will search for already existing matches
+*/
+export const biineSqlString = ({ table, records, unique }) => {
```
craigspaeth: 👍 this is a nice way to wrap more advanced raw SQL queries. Whenever dropping down to raw SQL remember of [SQL injection](https://www.w3schools.com/sql/sql_injection.asp) possibilities where if user input was sent into `table` here it could expose us to [ol' Bobby Tables](http://bobby-tables.com/).

This is not as problematic for protected background code, and GraphQL's type system helps us to some degree, but if this function were exposed via our API then we'd have to be more careful about it. In that case (and generally speaking) the best thing to do to is to sanitize the inputs via a [validation lib](https://github.com/hapijs/joi).

craigspaeth:
Lots of nit-picks you can feel free to ignore. The important comments are...

- https://github.com/findOctave/cortex/pull/679/files#r267013275
- https://github.com/findOctave/cortex/pull/679/files#r267065505
- https://github.com/findOctave/cortex/pull/679/files#r267066514

Maybe not necessary to address in this PR though, and we can discuss more IRL.

lionsdwarf:
PR revised for new `pg` schema where `is_notified` column on `person_salve` table replaces `person-salve-notification` table (RIP). 

Additionally, this single `sync-person` job started feeling leaky so there are now three separate sync jobs + a fourth "parent" job that calls each sync job in order: `pg-sync` calls `sync-person`, `sync-person-salve`, and finally `sync-notifications`.

We agreed to 🍐 through the "intermediate mapping" code and simplify/replace it with lodash logic--this work remains outstanding.

Also contained within is sanitation for raw SQL query as well as some new logging code. We should discuss the HIPAA implications for logging and will probably need to modify these output messages.

lionsdwarf:
@craigspaeth I will be modifying the table schema for `person_onboarding_notification` to mimic `person_salve`. The new table will be called `person_signup_stage` that will join `person` <> `lead.signupStage` with an `is_notified` column. 

These changes don't need to prevent a merge but just food for thought in your review.

PR #684, opened 2019-03-21
Use patient name/email for clients

craigspaeth:
A client's name was changed in Cliniko and it didn't reflect in their profile b/c it wasn't synced to Auth0. Auth0 actually doesn't support changing the name field, so this just drops using names in Auth0 and treats Cliniko as the source of truth for client name/email.


apps/api/models/client.js
```diff
@@ -100,7 +100,7 @@ export const createClientFromLead = async (_root, args) => {
   const lead = await Lead.findById(args.leadId)
   const [customer, patient] = await Promise.all([
     upsertCustomer({ email: lead.email, source: args.stripeSource }),
-    Patient.findById(lead.clinikoId)
+    Patient.findById(lead.patientId)
```
lionsdwarf: I'm confused--we don't have a `patientId` field on Lead, as I recall.

craigspaeth: Whoops! Good catch, that was a WIP edit that snuck in—will revert.

PR #688, opened 2019-03-22
Fix logo svg

craigspaeth:
The svg for the logo file was constructed of masks and overlapping shapes that caused weird issues like a pixel being shaved off the left side of it. I pulled it into Sketch, ungrouped, did some combine > subtract, so only the minimal geometry was left, then re-exported and it seemed to fix things.

image

Left: After; Right: Before


dGoligorsky:
💯💯💯

PR #697, opened 2019-03-27
Query for upcoming appts in Reschedule

lionsdwarf:
The /reschedule page was broken because it was trying to find an upcoming appointment in a list of past appointments.


craigspaeth:
Good catch! Thanks.

PR #713, opened 2019-03-31
Add timestamp or index for unique key per Zoho submission

craigspaeth:
I realized my assessment sync code would only sync once per assessment type when what we really want is once per submission. To do this I create a key based on the type's name and timestamp of submission to cross-check instead of just based on the type's name.


apps/api/jobs/etl/sync-assessments.js
```diff
@@ -10,36 +10,37 @@
...
 import * as airtable from 'at'
 import * as cliniko from 'cliniko'
 import * as log from '../log'
-import { google } from 'googleapis'
+import moment from 'moment'
 
-export const sheetId = assmtType => {
-  const split = assmtType.url.split('/')
+export const sheetId = assmntType => {
+  const split = assmntType.url.split('/')
```
lionsdwarf: I've been noticing "cannot read property split of 'undefined'" errors in our server logs. Maybe this was the culprit?

lionsdwarf: Yes, this was the culprit, I've confirmed from server logs. Do you want to add some defense ?

craigspaeth: Good catch, will update!

PR #715, opened 2019-03-31
PATCH cancellation instead of DELETE

craigspaeth:
A clinician was confused as to how an appointment disappeared. We were deleting appts. when we should be using Cliniko's [PATCH cancellation API](https://github.com/redguava/cliniko-api/blob/master/sections/individual_appointments.md#cancel-individual-appointment).


lib/cliniko/index.js
```diff
@@ -23,6 +23,18 @@ const throttle = new Throttle({
   concurrent: 120
 })
 
+export const path = ({
+  url,
+  resource,
+  id,
+  CLINIKO_API_URL = process.env.CLINIKO_API_URL
+}) => {
+  if (url && url.match(/^\//)) return `${CLINIKO_API_URL}${url}`
```
lionsdwarf: Should this have a slash like `${CLINIKO_API_URL}/${url}`?

lionsdwarf: If we use the `fs.path` module instead of string interpolation here, then we don't have to think about the slashes.

lionsdwarf:
Other than my path comment, LGTM!

lib/cliniko/index.js
craigspaeth: I'm trying to be explicit here about only concatenating if the url starts with a slash—therefore it wouldn't need to include the slash or accommodate paths without slashes. Maybe it'd be better to allow both `cliniko.find({ url: 'resource_name/foo' })` and `cliniko.find({ url: '/resource_name/foo' })` though?

Thinking it over more it might make sense to consolidate `resource` and `url` options into one `path` or `from` option that checks for the cliniko base url e.g. these would all work

```javascript
cliniko.find({ from: patient.treatment_notes }) 
cliniko.find({ from: 'patients' }) 
cliniko.update({ from: `/patients/${id}/cancel` }) 
```

Maybe that's a bit too magical though 🤔. Added that to [this ticket](https://trello.com/c/OxEN3AvR/255-%F0%9F%90%AC-refactoring-code-for-easier-maintenance-better-velocity) anyways.

In the meantime, I'll add at-least add tests for both cases to document. Appreciate the comment!

lionsdwarf: Ah, got it ! Thanks for the explanation. Feel free to merge when you're ready.

PR #719, opened 2019-04-01
Dynamic auth err msg link for missing accts

craigspaeth:
Hopefully helps clients who try to log in with an email they think they signed up with but haven't submitted their CC info yet.

image


lionsdwarf:
Looks good !

PR #736, opened 2019-04-05
Person patient

lionsdwarf:


apps/api/models/person.js
```diff
@@ -0,0 +1,43 @@
+import * as db from 'db'
```
craigspaeth: 👏 well done with this model.

apps/api/models/person.js
```diff
@@ -0,0 +1,43 @@
...
+export const upsert = async (column, person) => {
+  const colVal = { [column]: person[column] }
+  const existingPerson = await getByColVal(colVal)
+  return existingPerson ? db.knex('person')
+    .where(to(colVal))
+    .update(to(person))
+    : db.knex('person').insert(to(person))
+}
+
+export const getByColVal = async colVal => {
```
craigspaeth: This feels awkwardly named as it's sorta leaky to expose the underlying "column" nomeclature. Since this is basically forwarding args to `.where` could we just call this `where` as well? `Person.where({ leadId: ... })`

apps/api/models/person.js
```diff
@@ -0,0 +1,43 @@
...
+    .select('*')
+    .where(toColVal)
+    .map(from))[0]
+}
+
+export const findAll = () => db
+  .knex('person')
+  .map(from)
+
+export const from = ({
```
craigspaeth: 👍 good stuff. Now I'm gonna suggest the opposite of above 😛 Maybe being more explicit and calling this `fromPG` and `toPG` would be more obvious.

apps/api/models/person.js
```diff
@@ -0,0 +1,43 @@
...
+export const getByColVal = async colVal => {
+  const toColVal = to(colVal)
+  return (await db
+    .knex('person')
+    .select('*')
+    .where(toColVal)
+    .map(from))[0]
+}
+
+export const findAll = () => db
```
craigspaeth: 👍

apps/api/models/person.js
```diff
@@ -0,0 +1,43 @@
+import * as db from 'db'
+
+export const upsert = async (column, person) => {
```
craigspaeth: 👍

craigspaeth:
👏 really solid stuff here. My comments are just nitpicks.

apps/api/jobs/etl/sync-assessments.js
```diff
@@ -100,21 +102,25 @@ const fetchSheetData = spreadsheetId => {
...
 
-const syncAssmnt = async (assmnt, assmntTypeName) => {
+const syncAssmnt = async (assmnt, assmntTypeName, people) => {
   try {
     const [lead] = await airtable.find({
       table: 'leads',
       filter: `emailAddress = '${assmnt.email}'`
     })
     if (!lead) return log.msg(`Lead ${assmnt.email} doesn't exist`)
+    const patientId = _.find(people, { leadId: lead.id }).patientId
```
craigspaeth: 🤔 This highlights an awkward part of our system. What do we consider the source of truth for a person's email? Because right now we're going LeadByEmail > Person > Patient. My hunch is we should consider the patient as the source of truth for email like we're starting to do that [with client's email](https://github.com/findOctave/cortex/pull/684). In which case this job should actually just query a cliniko patient by email directly—meaning we could just delete code here.

I also think some of this could be extracted more into the model layer so we could call something like `Patient.createAssmtNot(assmt, assmntTypeName)` instead of all the logic being in the jobs.

Food for though, not blocking this PR.

apps/api/models/person.js
lionsdwarf: 👍  Agreed, this makes more sense

apps/api/models/person.js
lionsdwarf: 👍

lionsdwarf:
Updates incorporated to this PR per your suggestions. Unfortunately I'm unable to run the app locally and validate end-to-end, so let's hold off on merging this for now. 

Additionally, I see on Circle that we are still getting test failures due to missing DATABASE_URL env var...Odd...

Will try to figure out what's up w local build, as well as this env var issue....

PR #743, opened 2019-04-06
Send charge warnings for the day and delay actual charges

craigspaeth:
This slightly modified the charge job to warn of the day's upcoming charges and delay the actual charges from occurring by three hours. The hope being if any anomalies are warned about 3 hours will be enough time to kill a worker or something before damage occurs.

FYI: The env file update is to include `STRIPE_MAX_LIMIT`


apps/api/jobs/charge-sessions/index.js
```diff
@@ -12,23 +12,34 @@ import { isWorkingHours } from './validation'
...
   await Promise.all(
-    rollUpCharges(recentAppts).map(createCharge),
-    toChargeWarnings(rollUpCharges(upcommingAppts)).map(sendChargeWarning),
-    rollUpInvoices(recentAppts).map(sendInvoice)
+    // Warn of today's charges
+    toChargeWarnings(rollUpCharges(recentAppts)).map(sendChargeWarning),
+
+    // Send today's charges and missing CC notices 3 hours later
+    rollUpCharges(recentAppts).map(_.debounce(createCharge, threeHours)),
+    rollUpInvoices(recentAppts).map(_.debounce(sendInvoice, threeHours)),
```
craigspaeth: Debounce will return a copy of a function that can only be called after a millisecond interval. The functional nature of this code made it really easy to re-compose these to achieve the desired delay between sending.

lionsdwarf: Nice! I am familiar with debounce, having used it for for autocomplete input functionality in the past. Always nice to have a reminder!

lionsdwarf:
Glad this was so simple!

PR #744, opened 2019-04-07
Extract cache pattern into a lib and reuse for AT

craigspaeth:
This extracts the caching code inside the Cliniko lib into its own `cachingFetch` library that allows us to wrap any fetch including Airtable calls with the cache aside/refresh ahead patterns. If this sticks around we can consider swapping out the `store` variable with an actual caching service like Redis. For now, given our early stage/small traffic, I think caching in-memory should do the trick.


lionsdwarf:
Awesome--LGTM! 👍

PR #752, opened 2019-04-08
Add 3 min TTL to availabilities

craigspaeth:


craigspaeth:
Bleh test is still intermittent failing—I'll use Jest timer mocks to fix this.

PR #755, opened 2019-04-08
filter leads without ckoId

lionsdwarf:


apps/api/jobs/etl/sync-person.js
```diff
@@ -13,7 +13,9 @@ const syncPerson = async leads => {
   await db.bulkInsertIfNotExists({
     table: 'person',
     unique: ['lead_id', 'patient_id'],
-    records: leads.map(toPerson)
+    records: leads
```
craigspaeth: This would be an easy thing to test, but it's trivial enough I'll look past it 🙃

lionsdwarf: Just added a test.

PR #756, opened 2019-04-09
Save backups to PG

craigspaeth:
This refactors the backup job to save the backups to PG instead of zipping and emailing them via Mailgun.


README.md
```diff
@@ -8,13 +8,22 @@ The main Octave repo.
 
 Install the [latest version of Node](https://nodejs.org/en/).
 
+Install PostgreSQL, start the service, and create a database.
```
lionsdwarf: Thanks for taking care of this!

knexfile.js
```diff
@@ -1,3 +1,5 @@
+require('dotenv').config()
```
lionsdwarf: Forgot about this--good call!

apps/api/jobs/etl/backup.js
```diff
@@ -42,35 +40,23 @@ const backups = [
...
-    from: EMAIL_IT_ADDRESS,
-    to: EMAIL_IT_ADDRESS,
-    subject: `[Data Backup] From ${_.map(backups, 'service').join(', ')}`,
-    text: 'See attachments',
-    attachment
-  })
-  log.success(`Emailed backups to ${EMAIL_IT_ADDRESS}`)
+  await db
+    .knex('service_backup')
+    .del()
```
lionsdwarf: Looks like this wipes the table before insert--makes sense since Aptible backs-up daily (or more frequently). 👍

migrations/20190408180447_backups.js
```diff
@@ -0,0 +1,12 @@
+exports.up = function (knex, Promise) {
+  return knex.schema.createTable('service_backup', table => {
+    table.increments()
+    table.string('service').notNullable()
+    table.string('resource').notNullable()
+    table.json('data').notNullable()
```
lionsdwarf: For backups I agree that the `json` data type here makes most sense (as opposed to `jsonb` data type, which would give us more leverage for things like searching on the documents). 👍

PR #761, opened 2019-04-10
Replace debounce with sleep in charge job

craigspaeth:
This fixes the overly clever debounce thing I was doing and instead uses a sleep to delay the batch sending. 🤞 this has no conflicts with master.


craigspaeth:
Failure is from an intermittent issue [already fixed in master](https://github.com/findOctave/cortex/pull/754).

PR #762, opened 2019-04-10
Deploy sleep instead of debounce in charge job

craigspaeth:
PRing this to prod. so we can cherry-pick it to run tomorrow before moving forward with master.


craigspaeth:
Failure is from an intermittent issue [already fixed in master](https://github.com/findOctave/cortex/pull/754). Mergin'

lionsdwarf:
👍

PR #763, opened 2019-04-10
Sync person

lionsdwarf:
This is ready for prime-time.


apps/api/jobs/charge-sessions/network.js
```diff
@@ -139,11 +140,11 @@ const apptToAggregate = async ({
     billableItem,
     practitioner
   } = await fetchClinikoData({ appt })
-  // Ideally we'd fetch Auth0 by the patient's ID, but it's not possible.
```
craigspaeth: 🙌 so happy to see this gone

apps/api/jobs/charge-sessions/network.js
```diff
@@ -139,11 +140,11 @@ const apptToAggregate = async ({
...
-  // https://community.auth0.com/t/how-to-search-for-users-by-using-app-metadata/8688
-  const [authUser] = await auth0.getUsersByEmail(patient.email.toLowerCase())
+  const customerId =
+    patient &&
+    (await db.knex('person').where({ patient_id: patient.id }))[0].customer_id
```
craigspaeth: Non-blocker, but this could be an example of functions that can live on the model either generic like.. 

```javascript
Person.where({ patient_id: patient.id })[0].customer_id
```

but also don't be shy of composing tiny specific functions e.g.

```javascript
Person.firstCustomerId( // domain
  Person.where({ patient_id: patient.id }) // mapper
)

// or

Person.customerIdFromPatient(patient.id) // index.js composing mapper + domain
```

Although it might feel a little arduous, these tiny functions encourage reusability, easier/better test coverage, and the more we can hide direct database transactions the less coupled our entire codebase becomes to the implementation details of the data store.

Think of how Rails' ActiveRecord models hide direct database access even in layers outside the main MVC stack like background jobs or mailers. We should strive to achieve something like that separation. This includes access to third-party services like Stripe and Airtable. Obviously, this is feedback for both of us as I've littered the codebase with direct API/DB calls too, but it'd be nice to start getting more rigorous about it now that we're adding PG to the stack.

apps/api/jobs/etl/sync-person.js
```diff
@@ -1,35 +1,27 @@
...
-}
+const syncPeople = people => Promise.all(
+  people.map(person => Person.upsert(person))
+)
 
 export default async (leads) => {
   const people = await db.knex('person')
-  await syncPerson(people, leads)
+  const users = await auth0.users.getAll()
+  await syncPeople(updatePeople(people, leads, users))
```
craigspaeth: Nice composition of async and pure functions going on here.

apps/api/models/client.js
```diff
@@ -68,25 +68,19 @@ export const creditCardFromStripe = customer => {
...
   appointments: args =>
     Appointment.findForPatient({ patientId: patient.id, ...args }),
   salves: () => Salve.findAllForClient(patient.id)
 })
 
-export const toAuth0 = ({ lead, customer, patient }) => ({
+export const toAuth0 = ({ lead }) => ({
   email: lead.email,
   connection: 'email',
-  user_metadata: {
```
craigspaeth: 👏

apps/api/models/person.js
```diff
@@ -1,40 +1,74 @@
...
+  const colVal = { [columnProp]: person[columnProp] }
+  let existingPerson
+  if (!person.id) {
+    existingPerson = await findWhere(colVal)
+  }
+  return person.id || existingPerson ? updateWhere(person, colVal)
+    : insert(person)
+}
+
+export const findWhere = async where => (await pgWhere(toPG(where)).map(fromPG))[0]
```
craigspaeth: 👏 digging the composition of tiny functions.

craigspaeth:
Awesome! Great stuff 👏

apps/api/jobs/charge-sessions/network.js
lionsdwarf: I intentionally decoupled this from the model bc this is high-risk but I totally hear you!

craigspaeth: 💯 good call!

PR #765, opened 2019-04-10
[Do not merge] Prisma

lionsdwarf:
Auto-generated prisma config


lionsdwarf:
Marking this branch that attempts to "introspect" our existing psql schema as a timebox fail. I had to rewind to a different Prisma version in order to successfully `prisma deploy` and also made a number of weird config changes along the way to resolve issues. Now that the Prisma client is *not* working, I think there are too many variables at play to safely continue on this branch. I'm going to start anew with fresh psql instance.

PR #774, opened 2019-04-16
Audit Lead joins

lionsdwarf:
PSQL now stores all `Lead.signupStage` and `Lead.salves` joins.

Recap: `Lead.signupStage` lives within `person_signup_stage` table, which maps `Person.id` <> `Lead.signupStage`, + an additional `is_notified` column to track notifications.

`Lead.salves` now live in jsonb column `Person.salve`, containing object ```{ assigned: [], notified: [] }```

Also small addition to `log` lib for of nested job logging:
Screen Shot 2019-04-16 at 2 34 31 PM


lib/log/index.js
```diff
@@ -20,3 +20,27 @@ export const msg = msg => {
...
+      count--
+    }
+    return marker
+  },
+  init (job) {
+    console.log(`🤖 ${this.getMarker()}  Initialize job: ${job}...`)
+    this.jobCount++
+  },
+  fin (job) {
+    this.jobCount--
```
craigspaeth: This feels prone to race conditions, but maybe I'm overlooking something. e.g. imagine this scenario...

1. Job A `init`s `>> Initialize A`
2. Job B `init`s `>>>> Initialize B`
3. Job A `fin`s `>>>> Finalize A`
4. Job B `fin`s `>> Finalize B`

As you can see on step 3 job A looks like it's indented at the same level of job B.

In practice, our production log stream is going to be muddied with all sorts of async noise anyways so this `>>` indentation based way of looking at logs seems unideal. I'd support some kind of "tagging" of jobs so it's easier to `grep` against though e.g.

```javascript
log.msg('Hello world', 'charge-sessions') // 🤖 [charge-sessions] Hello world
```

It might also be worth looking at [prior npm art](https://github.com/winstonjs/winston)if we want to get fancy with logging. If this is intended mainly for local dev I might also suggest simply using `npm run web | grep some-string` to filter out noise.

Given this, I'm afraid I'm going to have to ask that you PR this log lib update separately seeing as it affects so many other places this lib is used.

migrations/20190416084253_create_person_salve_col.js
```diff
@@ -0,0 +1,12 @@
+
+exports.up = function (knex, Promise) {
+  return knex.schema.table('person', pt => {
+    pt.json('salve')
```
craigspaeth: Digging the use of JSON for these early/transitional fields. I wonder if we should lean into this sorta NoSQL approach more and have things like the `person_signup_stage` table become a `pt.json('signup_stage')` field that looks like:

```
signup_stage: {
  value: String
  is_notified: Boolean
}
``` 

Food for thought—not blocking for this PR.

apps/api/test/jobs/etl/sync-salves.test.js
```diff
@@ -0,0 +1,70 @@
...
+    }
+    const newNotified = sps.mergeNotified(lead)
+    expect(newNotified).toEqual(expect.arrayContaining([1, 2, 3]))
+    const lead2 = {
+      salves: [1, 2, 3],
+      receivedSalveNotifications: [1, 2, 3]
+    }
+    const notified1 = [1, 2]
+    const newNotified2 = sps.mergeNotified(lead2, notified1)
+    expect(newNotified2).toEqual(expect.arrayContaining([1, 2, 3]))
```
craigspaeth: `describe` and `expect.arrayContaining`... getting fancy with Jest I see!

apps/api/jobs/etl/sync-salves.js
```diff
@@ -0,0 +1,40 @@
+import * as db from 'db'
+import * as log from 'log'
+import _ from 'lodash'
+
+// Merge Lead.receivedSalveNotifications with Person.salve.notified. Person.salve.notified treated as
+// source of truth for diffs. Notified ids are removed if the salve is removed/deleted.
```
craigspaeth: Appreciate the comments, pure functions, and tests. This LGTM!

craigspaeth:
Mostly LGTM! Sorry to nit-pick but I'd rather you pull out the log library changes from this PR so we can discuss the best approach there separately.

After that's pulled out feel free to merge. I'm definitely noticing more places where it would feel better to have knex/API calls being made behind a data mapper layer in models, but not-block here. Once we get this syncing transitioning humming along in a solid place let's invest in an architectural cleanup effort that marches towards a cleaner post-Airtable PHI world in our code and database schemas.

lib/log/index.js
lionsdwarf: Good call. Shouldn't have tried to get fancy.

lionsdwarf:
Removed log lib change. 

I agree that there could be a signup_stage json column on Person--let's discuss priority to modify. 

Regarding the knex/API calls...I had originally intended to use direct knex callouts 
only within the to-one-day-be-abandoned sync jobs, but thanks for highlighting this because now I realize some knex bled into the notifications. :(  This is easy to clean up and I'll follow up with subsequent PR.

craigspaeth:
Thanks! No need to follow up with cleanups to notifications—there's plenty of other places to clean this up and I'd be happy to defer until after we start dropping columns in Airtable and have a cleaner slate to work with.

PR #776, opened 2019-04-17
Charge sessions prevent dupes

lionsdwarf:
Revised to use a json column. Will give us greater flexibility to store different kinds of data within this table without having to run migrations, should we wish to...


lionsdwarf:
Also uncommented our `isWokingHours` code...FYI

apps/api/jobs/charge-sessions/transaction.js
```diff
@@ -0,0 +1,35 @@
+import * as db from 'db'
```
craigspaeth: 👍 I like the code organization here. I wonder if it could even go one step further and be move the api/models and move `isDupeProcess` here as well as that's essentially business logic on the "status" model. Food for thought—not blocking.

craigspaeth:
Nice! LGTM!

PR #777, opened 2019-04-18
Json sync

lionsdwarf:
With `jsonb` columns for Person.`salve`, `onboard_stage`, and `coupon`, we now have a singular `sync-person` job that populates all json columns instead of populating hard join tables.  I think this is a good approach that will give us a lot of leverage in development speed going forward, and can also be easily transformed into relational join tables, if need be.

Also contains some cleanup to eliminate `db.knex` imports and refer directly to Person model instead.

As such, there is significant migration file cleanup and we will have to drop our Schema for deploy.


apps/api/models/person.js
```diff
@@ -1,14 +1,42 @@
...
+  const personOS = person && person.onboardStage
   const mergedPerson = {
     id: person && person.id,
     userId: user && user.user_id,
     leadId: lead.id,
-    patientId: lead.clinikoId || (person && person.patient_id),
-    customerId: (customer && customer.id) || (person && person.customer_id) || (user && user.user_metadata && user.user_metadata.customerId)
+    patientId: lead.clinikoId || (person && person.patientId),
+    customerId: (customer && customer.id) || (person && person.customerId) || (user && user.user_metadata && user.user_metadata.customerId),
+    salves: {
```
craigspaeth: 👍 nice, using JSON _does_ clean this up quite a bit.

craigspaeth:
👏 this is great. Hindsight is 20/20 and I'm a fan of how using the JSON type here simplifies things quite a bit. As you say, we can always migrate to more normalized data down the line but it's simpler to go JSON > tables than the other way around—especially for these transitional/early data structures.

PR #778, opened 2019-04-18
Terminate bulkInsert "with extreme prejudice"

lionsdwarf:
Remove "bulkInsert" code that is no longer needed now that we eliminated join tables


lionsdwarf:
Need to redeploy. Meging....

PR #781, opened 2019-04-19
Pg charge sessions

lionsdwarf:
Leverages `Person` model for: 
1) Reads in `jobs/charge-sessions`
2) Writes via `applyCouponToLead`

#2 illustrates how we probably want to strategize to move this and other Lead associations onto the Person model. Let me know if you want me to boy-scout that and update this PR, or whether we want to address Lead>Person associations wholesale....


apps/api/models/lead.js
```diff
@@ -208,14 +208,9 @@ export const applyCouponToLead = async (_root, args) => {
...
-  const updatedLead = await at.update({
-    table: 'leads',
-    fields: {
-      id: args.leadId,
-      couponCodes: updateCouponCodes({ lead, validCoupons })
-    }
-  })
-  return fromAirtable(updatedLead)
+  const person = await Person.findOne({ leadId: args.leadId })
+  await Person.applyCoupon(person, validCoupons[0].id)
```
craigspaeth: Nice 👍 totally agree that `applyCouponToLead` will move outside of lead. We don't have to do that in this PR but we'll definitely be heading that direction where Person will start to absorb things we've been modeling under Client and Lead.

Alternatively, we might consider a coupon as living under the "client" domain model... but practically speaking it might simpler to move towards a world that splits client out into "user" and "person" that aligns with the backing data stores (former being Auth0, later being our own database).

apps/api/models/status.js
```diff
@@ -0,0 +1,22 @@
...
+export const findByProcess = async processName =>
+  (await db.knex('status').where({ process: processName }))[0]
+
+export const update = (processName, data) =>
+  pgUpdate(processName, JSON.stringify(data))
+
+export const insert = (processName, data) =>
+  pgInsert(processName, JSON.stringify(data))
+
+const pgUpdate = (processName, data) =>
```
craigspaeth: 👍 appreciate the status modeling here.

apps/api/models/lead.js
lionsdwarf: 👍

craigspaeth:
👏 great stuff! Let's defer the [Scouts](https://world.wng.org/content/boy_scouts_of_america_announce_new_gender_neutral_name) rule on this one for now to keep the diff small and give us another chance to consider where we want to draw the domain lines around leads/clients/person when we're closer to a world where the leads table is getting killed and the data is in a cleaner non-syncing state.

PR #782, opened 2019-04-20
[Do not merge] Toggle Salve/Coupon assignments

lionsdwarf:
On the plane today I built a little react component for an admin to assign Salves and/or Coupons to a Person. Will need to be wired up to a "Select Person" component, and still no auth or anything like that yet--just stubbed data.

assign


apps/manage/components/AssignToggle.js
```diff
@@ -0,0 +1,52 @@
...
+    return (
+      <div>
+        <h3>{this.props.assignType}</h3>
+        {
+          this.state.assignable.map(card => <Card 
+            {...card}
+            onClick={() => this.setState(toggle(this.state, card))}
+          />)
+        }  
+        <div>~~~~</div>
```
craigspaeth: 👏 this is a great start. I'm not going to be too picky about the UI/UX here but if it wouldn't be too much more work I think checkboxes would be a more intuitive UI. That might look something like...

```javascript
<ul>
  {this.state.assignable.concat(this.state.assigned).map(coupon => (
    <li>
      <input
        type='checkbox'
        checked={_.find(this.state.assigned, { id: coupon.id })}
        onClick={() =>
          this.setState({ assigned: [...this.state.assigned, coupon] })
        }
      />
    </li>
  ))}
</ul>
```

Also, keep in mind this will save to the server so you might not even need local state to keep track of the checked boxes as a page refresh should load them correctly and you can rely on the default HTML behavior to toggle the ✔️ (but don't quote me on that... I can't remember offhand if React prevents this).

apps/manage/components/Cards.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react'
+
+export const CouponCard = ({ code, discount, onClick }) => <div
```
craigspaeth: Nitpick: Card is a slightly odd name for this as card to me implies the [Material Design card](https://material.io/design/components/cards.html) of an image + color block + some copy + action button. "List item" or `li` might be more appropriate.

craigspaeth:
Code looks good! I have some nitpicks but ultimately I'm happy to merge this for now. I _do_ think you could simplify this from a UX perspective by using checkboxes, and I wonder if that could result in a more intuitive breakdown of components so that there's only a `AssignCheckboxes` instead of passing a `card` into the AssignToggle component.

Feel free to pursue that route and update the PR or feel free to merge when 💚 if you'd rather move forward. Thanks!

PR #785, opened 2019-04-23
terminate bad code "with extreme prejudice"

lionsdwarf:
Removes some old artifacts that should never have made it into master


craigspaeth:
LGTM! Feel free to merge when 💚

PR #792, opened 2019-04-24
Cs fix

lionsdwarf:
Paginate requests for all Auth0 users


apps/api/models/user.js
```diff
@@ -0,0 +1,24 @@
...
+export const findAll = (
+  options = {
+    page: 0,
+    per_page: AUTH0_PAGINATION_LIMIT
+  },
+  cummulativeUsers = []
+) =>
+  auth0.users.getAll(options).then(res => {
+    if (res.length === options.per_page) {
+      options.page++
```
craigspaeth: I don't _think_ this is problematic but it might be better to use an immutable approach here `return findAll({ ...options, page: options.page + 1 })`

craigspaeth:
LGTM! Feel to merge when 💚

apps/api/models/user.js
lionsdwarf: Good call. I've revised this.

PR #793, opened 2019-04-24
signupStage fix

lionsdwarf:
`Lead.signupStage` gets converted to upper snake case in `Lead.findById`, and it was not getting converted back to camel case before `Person.merge`


apps/api/models/client.js
```diff
@@ -98,6 +99,7 @@ export const createClientFromLead = async (_root, args) => {
     Patient.findById(person.patientId)
   ])
   const user = await auth0.createUser(toAuth0({ lead }))
+  lead.signupStage = _.invert(Lead.signupStageMap)[lead.signupStage]
```
craigspaeth: This seems like it should live under `Person.upsert` or another person model function with a test if that's why we added this bit.

craigspaeth: 👍 works for me, but making note this would be a good place to think about our separation of concerns between IO transactions like this upsert function and biz-logic/data-shaping functions that would handle this conversion. Feels a bit out of place here and should ideally have a test with it.

PR #794, opened 2019-04-25
fix empty signupStage for class bookings for new patient

lionsdwarf:
Guard against new patients lacking signupStage (for class bookings)


lionsdwarf:
Merging so that I can finish QA > Deploy...

apps/api/models/lead.js
```diff
@@ -84,23 +84,30 @@ export const fromAirtable = record => ({
   appointments: () => Appointment.findForLead(record)
 })
 
+const joinSalvesAndOnboardStage = async lead => {
```
craigspaeth: 👍 nice refactor

PR #796, opened 2019-04-25
[ WIP - do not merge ] Sync Person.name/email, Search for Person by name/email

lionsdwarf:
Adds `name` and `email` columns to `person` table. Decided to add hard string columns instead of json since: 1) we are searching directly on these columns, 2) so far our pattern for using json data type is for joins. Don't think we lose much by making these string type columns.

Frontend-wise, using react-select lib to search on these columns.

You can ignore hard-coded Salve and Coupon data in screencap--working on making that dynamic... Jumping into a `Content` model now that will pull from AT. As I dig in here I'm starting to question the hastily designed schema for Coupons and OnboardStage which expose `assigned`, `notified` fields. Seems like these associations should live on the Person directly, and each type should be revised with fields like `id` and, for Coupons, `code`, etc.

Also, lots of file structure/design still in flux so take that with a grain of salt. Also, I've thus-far resisted the temptation to break the Person model down using the data-mapper pattern--I think it makes sense to address wholesale with a well-defined strategy rather than diving in ad hoc (we have a ticket for this).

Let me know if you see any glaring red flags, otherwise I shall press on...

search_person


apps/api/models/person.js
```diff
@@ -37,11 +77,15 @@ export const merge = (lead, person, user, customer) => {
       current: lead.signupStage,
       assigned: mergeArrs(personOS && personOS.assigned, [lead.signupStage]),
       notified: mergeArrs(personOS && personOS.notified, lead.receivedNotifications)
-    }
+    },
+    name: lead.name,
```
craigspaeth: This should ideally `patient` as people will expect Cliniko to be the source of truth for these fields. For now, this is fine b/c it's all syncing every-which-way but note for the future.

craigspaeth:
👍 looking good! The only thing that comes to mind is how this introduces duplicated data (name + email) between Lead/Cliniko and our DB. I wonder if we'll want to come up with a more explicit convention for that other than copying over the ["scalar"](https://softwareengineering.stackexchange.com/questions/238033/what-does-it-mean-when-data-is-scalar) fields on an as-needed basis—like use a `patient_snapshot` JSON field or separate `patient_snapshot` table with a `person_id` column. I don't know the right answer, but it seems unobvious that `name` and `email` in our own database are supposed to be "replica" fields

PR #797, opened 2019-04-26
Salve log API

craigspaeth:
This adds an API over "salve results" so that we can use salves for logs. Salve results are basically the spreadsheet data from a Zoho form. As such this PR comes with a couple changes to support that:

- Adds a `resultsUrl` to salves to store the google sheet (pictured below)
- Refactored out some shared G Sheets operations into an apps/api/lib/sheets lib
- Added GQL type and query to the client model to query for `salveResults` (already authorized b/c it's a child of the client query—pictured below)
- Added a placeholder page component that renders the raw data (pictured below)

image

cap

image


apps/api/test/models/salve.test.js
```diff
@@ -97,3 +97,51 @@ test('orderedSalves returns salves in the related record order', () => {
...
+        anxious: true
+      },
+      {
+        email: 'karen@h.com',
+        anxious: true
+      }
+    ])
+  ).toEqual([
+    {
+      columns: [
```
lionsdwarf: Semantically-speaking, seems to me like this is  a `row`?

apps/api/test/models/salve.test.js
```diff
@@ -97,3 +97,51 @@ test('orderedSalves returns salves in the related record order', () => {
...
+        {
+          key: 'email',
```
lionsdwarf: ...and "key" is a `column`.

apps/api/models/salve.js
```diff
@@ -1,10 +1,18 @@
...
 import * as Person from './person'
-import _ from 'lodash'
+import * as sheets from '../lib/sheets'
 
 export const schema = {
   types: `
+    type SalveResult {
+      columns: [SalveResultColumn]
+    }
+    type SalveResultColumn {
```
lionsdwarf: Maybe I'm missing something here, but again, this seems like a row, and the key would be the column.

lionsdwarf:
I made and then deleted some comments about row vs column semantics. Now I understand that we are dealing with dynamic collections of columns, which explains why things are broken down by column instead of row. So if you saw my prior, confused comments, please ignore. 

Having said that, this LGTM! Merginnnnn

craigspaeth:
np! Appreciate the comments. Yeah, I was struggling to come up with more obvious data modeling/naming 🤔 I might refactor that later.

PR #798, opened 2019-04-27
Salve log UI

craigspaeth:
Adds the UI to the salve log results. I also slightly refactored the API to look like...

```
SalveResult {
  submittedTime: UTCString
  questions: [
    {
      name: String
      answer: String
    }
  ]
}
```

... to be a bit closer to how we think about the domain than modeling a spreadsheet.

Uses the [rechart](https://github.com/recharts/recharts) library:

Screen Shot 2019-04-26 at 11 11 03 PM


apps/api/models/salve.js
```diff
@@ -3,15 +3,17 @@ import { airtableImgToResizeUrl } from './content-module'
...
   types: `
     type SalveResult {
-      columns: [SalveResultColumn]
+      submittedTime: String
+      questions: [SalveResultQuestion]
     }
-    type SalveResultColumn {
-      key: String
-      value: String
+    type SalveResultQuestion {
```
lionsdwarf: Like this renaming here 👍

lionsdwarf:
This looks good! Nice that the chart library looks like it was pretty simple to use. Only one concern here...from the charts you posted, it looks like you're waking up a lot at night. :/

;)

PR #800, opened 2019-04-29
Revise document format for jsonb columns in Person table

lionsdwarf:
JSONB columns for `salve`, `onboard_stage`, and `coupon` previously stored ids in the relational format returned by Airtable, eg `{ assigned: [ id ], notified: [id] }`. In order to eschew a tangled data-mapping web for pg read/writes, we now adhere to a revised document structure for these columns as follows:

`coupon`: [ { id } ]
`salve`: `[ { id, isNotified  } ]` 
`onboard_stage`: `{ name, isNotified}`


apps/api/jobs/notifications/onboard-stage.js
```diff
@@ -19,13 +19,13 @@ export const isNotificationPeriod = time => {
...
-const sendNotification = async ({ content, lead, person, signupStage }) => {
+const sendNotification = async ({ content, lead, person, onboardStage }) => {
   if (process.env.IS_NOTIFICATIONS_ACTIVE === 'true') {
     await Person.update({
       id: person.id,
       onboardStage: {
-        ...person.onboardStage,
-        notified: Person.mergeArrs(person.onboardStage.notified, [signupStage])
+        name: onboardStage,
+        isNotified: true
```
craigspaeth: Small thing to note: by changing this from `receivedNotifications: [ID]` to a single map this will mean that someone who changes from `referral` to `schedulingTherapy` and then back to `referral` will receive the referral email twice... but I'm going to say that's actually a _feature not a bug_ and better than the list of historical `receivedNotifications` states we had before so 👍 just wanted to point that behavior change out.

apps/api/models/person.js
```diff
@@ -92,6 +114,18 @@ const pgWhere = async where => db
 const pgAll = () => db
   .knex('person')
 
+export const oldSalvesToArray = salves => salves.assigned.map(salveId => ({
+  id: salveId,
+  isNotified: salves.notified.includes(salveId)
+}))
+
+export const oldCouponsToArray = coupons => coupons.assigned.map(id => ({ id }))
```
craigspaeth: I'm cool with this code for now b/c we're digging ourselves out of this in-between Airtable/PG state, but just FYI that this is kinda an antipattern to gradually migrate old schemas like this in application code. We'd ideally write a migration script that makes this update wholesale.

At least, that's what the RDBMS folks would say. Ask [a NoSQLer](https://derickrethans.nl/managing-schema-changes.html) and they might consider this one of various legit ways to support data modeling evolution.

craigspaeth:
Only nitpicks/non-blocking comments. LGTM! Want to sync up IRL about billing though.

apps/api/test/models/person.test.js
```diff
@@ -361,22 +464,40 @@ describe('findToNotify', () => {
...
+        { id: 2, isNotified: true },
+        { id: 3, isNotified: false }
+      ])
+    )
   })
-  test('calls "where" with id as person id', () => {
-    expect(db.knex().where).toHaveBeenCalledWith({ id: '1p' })
+})
+
+describe('oldOnboardStageToObj', () => {
```
craigspaeth: 👏 Great job organizing the code into testable pieces + great test coverage!

PR #801, opened 2019-04-29
Support log types in cards

craigspaeth:
This adds a couple small features to support log type salves better:

- Injects the client's email to prefill a Zoho form within the iframe
- Support a new icon and "X min survey" copy
- Callout pill in the corner of a log linking to the log results page

Screen Shot 2019-04-29 at 6 27 27 PM


apps/api/models/appointment.js
```diff
@@ -86,12 +86,16 @@ export const rateFromCliniko = async apptType => {
   return { amount: item.price }
 }
 
-export const isClass = apptType => apptType.category.match(/class/i)
+export const category = (apptType) => {
+  if (!apptType.category) return null
```
craigspaeth: Sorry I forgot to branch and accidentally snuck in a small bug fix to gracefully handle appointment types that are missing categories. (e.g. Mego has updated category to "Group" and emptied it and it's taken the class page down b/c we assumed it was always going to be something in our enum.

lib/auth/index.css
```diff
@@ -126,7 +126,7 @@
   }
 
   .auth0-lock-overlay {
-    background: rgba(0, 0, 0, 0.7) !important;
+    background: var(--dark-scrim) !important;
```
lionsdwarf: oooh  nice.

lionsdwarf:
LGTM!

PR #803, opened 2019-05-01
Basic embedded careers page

craigspaeth:
This embeds our Greenhouse hosted career page and hosts a CSS file we can use to get basic typography matching by pointing to it in Greenhouse.

image


apps/marketing/pages/careers.js
```diff
@@ -0,0 +1,48 @@
+import { GraphQLClient } from 'graphql-request'
```
craigspaeth: Apologies for the linter noise—I had to update prettier to be able to ignore the hosted CSS file (it was choking on the base64 encoded font). This is the only important file.

lionsdwarf: Thanks for the heads up.

lionsdwarf:
Easy peasy ! Merginnnn...

PR #804, opened 2019-05-01
jobs use job-specific retrieval methods on person model

lionsdwarf:
There is some duplicate code in here for the moment, which will get resolved in my current working branch.

Also picked up some linter noise. You can ignore anything not in `apps/api/jobs`.


craigspaeth:
👍 feels pretty explicit and works for me! Thanks!

PR #808, opened 2019-05-02
Fix CSS of salve cards

craigspaeth:
Small CSS tweaks primarily to fix the margin-top error, but also tiny design fixes for better visual harmony.

image


craigspaeth:
Margin' and deployin'

PR #812, opened 2019-05-03
More direct assessment sync

craigspaeth:
I was investigating a reported assessment note syncing issue and while I didn't find anything off in the code I decided it'd be good to do a _scouts rule_ and clean up the file to skip the Lead > Person > Patient query and just directly query patient by email (seeing as patient is our "source of truth" for email I think this makes sense). Down the line, it might make sense to move more of this logic to the `Patient` model or an `Assessment` model, but this seemed like pretty low hanging fruit worth implementing.


lionsdwarf:
👍

PR #817, opened 2019-05-04
Send salve result notifications to practitioners

craigspaeth:
This implements a way for clinicians to see their clients results easily by emailing them when a new result is submitted.

It does this by polling the results spreadsheet at an hourly interval and finding any results that were submitted within the last hour. It then finds the client by the email of the result, rolls up the data with Cliniko data like who the associated practitioner is, and sends an email with a link to the /content-results page.

Also included:
- Added some divider comments in models to better organize and start signifying where we should break up functions b/t domain/mapper/resolver/schema
- Minor env var cleanup

image


lionsdwarf:
Once the merge conflict is resolved, feel free to merge!

PR #820, opened 2019-05-06
Minor class page updates

craigspaeth:
1. Lowers cache TTL on Cliniko appointment types so the updates surface faster
2. Removes the second "Our Approach to Mindfulness" section


apps/api/jobs/charge-sessions/index.js
```diff
@@ -53,5 +53,5 @@ export default async () => {
   })
 
   // Close the db connection to avoid the cron from hanging
-  await db.knex.destroy()
+  await db.knex().destroy()
```
craigspaeth: Snuck this typo fix in... okay score one for type saftey here 🙃

craigspaeth:
Going to merge and deploy as these are pretty harmless/time sensitive.

PR #826, opened 2019-05-07
Show time types from Zoho in results page

craigspaeth:
Two things were preventing "times" from showing as graphs correctly in the results page:
1. We were cleaning out any falsy or "empty" values in G Sheets results which was causing that question to not show at all on the results page when rolling it up into graphs. (This was semi-useful for shorter looking treatment notes, but I'm just removing it to render the full question/answer digest in both places)
2. The data value looks like a `H:MM` format string which doesn't get cast into a number using just `Number` so I used separate rollup `filter` and `cast` functions to match against that data


apps/portal/components/content-results/index.js
```diff
@@ -92,7 +92,17 @@ export default ({ salve, results, client }) => (
           <Chart xKey='dateLabel' yKey='answer' data={data} />
         </div>
       ))}
-      {dataByQuestion(results, a => !Number(a), String).map(
+      {dataByQuestion(
+        results,
+        a => a && a.match(/[0-9]\:[0-9][0-9]$/),
```
craigspaeth: This `dataByQuestion` uses this `filter` function to filter out results in the results rollup, so we use a regex here to match against the `H:MM` string type and filter down to those results.

apps/portal/components/content-results/index.js
```diff
@@ -92,7 +92,17 @@ export default ({ salve, results, client }) => (
...
+        a => Number(a.split(':')[0]) + (a.split(':')[1] ? .5 : 0)
```
craigspaeth: This is the `cast` function used in `dataByQuestion` which "typecasts" to a numerical number that the charts library can use. In this case we're parsing out and cast the hour to a number then adding a `.5` depending on if there's a non-zero value in the `:MM` part.

lionsdwarf: Copy that 😬

lionsdwarf:
Looks good, Craig!

PR #829, opened 2019-05-09
UI update status

lionsdwarf:
assign_fail
assign_succes


craigspaeth:
Looking good. Happy to sync up about this in realtime soon!

lionsdwarf:
@craigspaeth This has been updated in the following ways: 

1) UI spinner state during async requests is now managed using local state and `useRefs` to track and compare updates to global state 
2) `assign' view-model action composition refactored and simplified
3) small bug fix to resolve Onboard Stage deletes failing

apps/manage/components/Assign.js
```diff
@@ -1,21 +1,47 @@
-import React from 'react'
+import React, { useEffect, useState, useRef } from 'react'
 import { useAppState } from '../lib/view-model'
 import AssignToggle from './AssignToggle'
 import Select from 'react-select'
 import _ from 'lodash'
 
+// Dives spinner by storing previous "updateStatus" for comparison of updateStatus
+const usePrevious = (value) => {
+  const ref = useRef()
```
craigspaeth: This is an odd use of refs. Refs are really more meant for dropping down to interact with DOM elements directly when needed and not as containers for state history. I think there's possibly a more straightforward way to do this that I'll comment below.

apps/manage/lib/view-model/assign.js
```diff
@@ -165,8 +166,36 @@ export const sortItems = (state, assigned, type, oldIndex, newIndex) => ({
...
+    }
+    : {
+      ...state,
+      ...initialState
+    }
 
+export const handleError = async req => {
+  let res
+  try {
+    res = await req()
```
craigspaeth: You could just `return req()` here instead of hoisting the value into `let res`

apps/manage/components/Assign.js
```diff
@@ -52,9 +84,16 @@ export default ({
             isSortable
           />
           <button
-            onClick={updatePerson}
-            disabled={_.isEmpty(state.assignments)}
+            onClick={() => {
+              setState({ isSpinner: true })
+              updatePerson()
```
craigspaeth: `updatePerson` is an async function, so I think you could simply model the state transitions right inside this click handler....

```javascript
onClick={async () => {
  setState({ isSpinner: true })
  await updatePerson()
  setState({ isSpinner: false })
  await sleep(4000)
  clearStatusState()
}}
```

This way you don't have to do the [useEffect](https://github.com/findOctave/cortex/pull/829/files#diff-b4a604a1d680b1766affe12dbb5246d7R31) or [ref](https://github.com/findOctave/cortex/pull/829/files#diff-b4a604a1d680b1766affe12dbb5246d7R9) things above.

craigspaeth:
Getting closer but I think you can do this a lot more straightforwardly. Try [my last comment](https://github.com/findOctave/cortex/pull/829#discussion_r282882581) and lmk how that works.

apps/manage/lib/view-model/assign.js
lionsdwarf: 👍

craigspaeth:
Thanks for the iterations, looks great!

PR #831, opened 2019-05-09
Hide results pill when there are no results for the client

craigspaeth:
Sorry this is kinda noisy for a small thing. A lot of that noise is just passing the salve results down the React chain to the `isResultsPillVisible` which is the thing doing the work here. This makes sure the `Recent results` pill button doesn't show if the client hasn't submitted any results yet.


lionsdwarf:
👍

PR #832, opened 2019-05-10
Guard against no assigned salve logs

craigspaeth:
Due to my latest PR the profile page will b0rk if someone doesn't have any salve logs assigned b/c it needed guard statements. Two things here:

1. Was getting back `null` from the API level I believe b/c of person salve sync-hole related discrepancies (might be only reproable on dev, but decided to `compact` those out anyways)
2. On the profile page there needed to be a conditional fetching of salves if none of the assigned salves were logs


apps/api/jobs/cli.js
```diff
@@ -4,4 +4,7 @@
  */
 import * as log from 'log'
 
-require(`./${process.argv[2]}`).default().catch(log.error)
+require(`./${process.argv[2]}`)
+  .default()
+  .then(() => process.exit())
```
craigspaeth: When running `sync-person` I noticed an opportunity to make this hanging process b/c of DB connection simpler to handle by simply making sure all jobs exit when they're finished via the `npm run job` script.

lionsdwarf: Nice!

PR #836, opened 2019-05-10
Style up admin panel

craigspaeth:
Adds some CSS and a little rejiggering of HTML and state to get this UI:

cap


apps/manage/lib/view-model/assign.js
```diff
@@ -166,28 +167,29 @@ export const sortItems = (state, assigned, type, oldIndex, newIndex) => ({
...
-  }
-  : state
-
-export const setUpdateStatus = (state, res) => res && res.id
-  ? {
-    ...state,
-    ...initialState,
-    updateStatus: {
-      isSuccess: true
+export const resetStateOnSuccess = state =>
```
lionsdwarf: Nitpick: this function name is now somewhat of a misnomer since it's just updating that one prop, and we already have a `setUpdateStatus` function. Maybe this logic should be moved back into `setUpdateStatus`, or there could be separate functions for each kind of update.

lionsdwarf:
Oh....and it looks boss! Nice work 👏

PR #837, opened 2019-05-10
Subway styles

lionsdwarf:
MTA line colors according to: http://web.mta.info/developers/resources/line_colors.htm

Screen Shot 2019-05-10 at 4 21 54 PM


lib/styles/index.css
```diff
@@ -13,6 +13,10 @@
   --white-color: #fff;
   --yellow-color: #eb991b;
   --dark-scrim: #232323ba;
+  /* MTA subway colors: http://web.mta.info/developers/resources/line_colors.htm*/
+  --subway-456: #00933c;
```
craigspaeth: Seeing as these probably won't be used outside of the locations component can we move these vars inside there?

craigspaeth:
LGTM but I'd prefer the vars not be in the global colors

lib/styles/index.css
lionsdwarf: Oh yea, sorry, good call

PR #838, opened 2019-05-10
Nav links toggle

lionsdwarf:
nav_toggle


lionsdwarf:
There is some linter noise. Only the CSS file has changes.

dGoligorsky:
🔥 Thanks for making that fix. I'll merge!

PR #840, opened 2019-05-13
Content results page QA

craigspaeth:
Two small bugs on the salve results page:

1. Questions with "empty" names (spreadsheet wackiness 🤷‍♂) were showing up which were confusing so filtering those out on the API side
2. To get the X and Y axis sorting properly I had to jump through some hoops with rechart to plot based on a numerical value but customize the labels via `formatter` functions


lionsdwarf:
LGTM !

PR #835, opened 2019-05-10
[ WIP -- DO NOT MERGE ] Charge sessions errors

lionsdwarf:

PR #842, opened 2019-05-14
Capture missing data errors better in charge job

craigspaeth:
This is a bit more surgical at capturing the missing patient/practitioner data in the charge job. It simply catches 404 errors and re-throws them as something easier to read. It also better handles the promises in the root function of the charge job and adds a couple more logs in there.


apps/api/jobs/charge-sessions/network.js
```diff
@@ -51,10 +56,12 @@ const fetchClinikoData = async ({ appt }) => {
...
-  const [practitioner, patient] = await Promise.all([
-    cliniko.find({ url: appt.practitioner.links.self }).catch(() => null),
-    cliniko.find({ url: appt.patient.links.self }).catch(() => null)
-  ])
+  const practitioner = await cliniko
+    .findOne({ url: appt.practitioner.links.self })
+    .catch(rethrowMissing('practitioner', appt))
+  const patient = await cliniko
+    .findOne({ url: appt.patient.links.self })
+    .catch(rethrowMissing('patient', appt))
```
craigspaeth: One thing I noticed [in your PR](https://github.com/findOctave/cortex/pull/835/files#diff-bf21562e83715209350088372227d9b9R58) is you were throwing after the `.find` call which it wouldn't reach because Cliniko doesn't return null in the case of a missing resource—they'll just throw a 404. So we have to explicitly catch that in the fetch here.

lionsdwarf: Ah gotcha! Thanks for the update. 

You're probably right that at some point we should hoist that 404 logic to the Cko lib

lionsdwarf:
LGTM! Thanks for tacking this.

PR #846, opened 2019-05-14
Kill onboard stage

lionsdwarf:
Removes all references to "signup stage" and "onboard stage". Includes the following changes:
1) New `notify` `api/lib` containing the `sendEmail` and `sendSMS` functions
2) Referral email notification fires from `createLead` resolver


craigspaeth:
🔥 🔥 🔥 🔥 

Love the −597 diff! LGTM feel free to merge when 💚

apps/api/models/lead.js
```diff
@@ -181,32 +169,32 @@ export const upsert = async fields => {
...
   return lead
 }
 
+export const sendReferralNotification = async lead => {
+  const content = (await at.find({
+    table: 'notifications',
+    filter: `name = 'referral'`
+  }))[0]
+  sendEmail(lead, content).catch(log.error)
+  sendSMS(lead, content).catch(log.error)
```
craigspaeth: 👍 nice use of `catch` and skipping the `await`s here to make sure we don't block or throw a lead creation if the referral email happens to not succeed.

lionsdwarf: 👍

PR #848, opened 2019-05-15
Sync practitioner to users script

craigspaeth:
This takes a humble first stab at syncing Cliniko practitioners as admins by simply finding practitioners in Cliniko and creating Auth0 users for them. I'm punting on the role upgrade for now as it seems like more trouble than it's worth right now to do that programmatically, so I'll just do it manually through the UI.


apps/api/jobs/etl/sync-practitioners.js
```diff
@@ -0,0 +1,14 @@
...
+ */
+import _ from 'lodash'
+import * as Client from '../../models/client'
+import * as Practitioner from '../../models/practitioner'
+import * as log from 'log'
+
+export default async () => {
+  const emails = _.map(await Practitioner.findAllUsers(), 'email')
+  log.msg(`Syncing ${emails.length} as admins..`)
+  await Client.createUsers(emails)
```
lionsdwarf: Feels a little wrong here to use the `Client` model to create practitioner users.

apps/api/models/client.js
```diff
@@ -176,3 +176,6 @@ export const findByEmail = async email => {
   const user = await auth0.getUser({ id: person.userId })
   return user
 }
+
+export const createUsers = async emails =>
```
lionsdwarf: Wouldn't it make more sense for this to live on a different model, eg `Users.create` ?

apps/api/models/practitioner.js
```diff
@@ -151,6 +151,16 @@ export const confirmIsAvailable = async ({
   )
 }
 
+export const findAllUsers = async () => {
```
lionsdwarf: Nitpick: This was a little confusing because of the naming collision w Auth0 Users. Maybe `findAllClinikoUsers` is more clear?

apps/api/jobs/etl/sync-practitioners.js
craigspaeth: 👍 good point didn't realize we had a user model. I'll move this stuff there. 🤔 might make sense to move more auth0 related stuff inside client to that user model too.

lionsdwarf: Agreed! Feel free to boy scout that.

PR #851, opened 2019-05-16
Eliminate all references to old json documents

lionsdwarf:
Found some lingering artifacts that were overlooked in yesterday's purge of support for old json documents. Also, some of yesterday's deletions got re-introduced to master somehow (in case any of this looks familiar...).

I also removed support for a single-assign-toggle since we're not using that since abandoning Onboard Stage.


craigspaeth:
❤️ some serious spring cleaning going on here!

PR #854, opened 2019-05-16
[WIP Do not merge] Ooo model breakup

lionsdwarf:
This code spike provides a Person class for use across b̶a̶c̶k̶g̶r̶o̶u̶n̶d̶ backend processes and includes a `toGQL` method on the class for use within resolver functions. I also split up the model code as inspired by the DataMapper pattern.

I think this provides solid separation of concerns, particularly in it's decoupling of application code and the database schema. It also eschews hierarchical inheritance patterns.


lionsdwarf:
I should add that the scope of these changes relate to:
1) A functioning `/assign` page
2) A (semi-) functioning `notifications/assigned-salves`
If you try to connect the changes beyond #1 & #2 (eg `sync-person`), you will find that the code breaks

apps/api/models/person/domain.js
```diff
@@ -0,0 +1,114 @@
...
+    this.coupons = pgPerson.coupon
+    this.name = pgPerson.name
+    this.email = pgPerson.email
+  }
+
+  toGQL = () => {
+    const salves = this.salves
+    const coupons = this.coupons
+    this.salves = () => Salve.findForPerson(salves)
+    this.coupons = () => Coupon.findForPerson(coupons)
```
craigspaeth: You'll want to be careful about mutating the same property on an instance like this. Notice in the constructor you assign `this.coupons =` as static data, then later here you turn it into a function. This means later in the code if someone tries to use `person.coupons` as data [like here](https://github.com/findOctave/cortex/pull/854/files#diff-d9d3223da3abf2053217e838f9d552caR29) it will introduce a subtle bug that throws an error or worse continues on with some data map that tries to calculate coupons as `{ coupons: Function }`.

Typically in OO you want to separate data from methods and not intermix the two, so that `this.coupons` is reserved for data and `this.getCoupons` or as some frameworks do it `this.get('coupons')` returns that data.

apps/api/models/person/domain.js
```diff
@@ -0,0 +1,114 @@
...
+    return this
```
craigspaeth: I think this would work fine, but I believe GraphQL would prefer more vanilla JSON-like data objects. `this` is going to include prototype information and have un-obvious interactions with `JSON.stringify`. I wonder if we would want to consider mixing in or extending a base class that provides some `toJSON` convention that `toGQL` would wrap.

apps/api/models/person/domain.js
```diff
@@ -0,0 +1,114 @@
...
+  constructor(pgPerson) {
+    this.id = pgPerson.id
+    this.leadId = pgPerson.lead_id
+    this.patientId = pgPerson.patient_id
+    this.customerId = pgPerson.customer_id
+    this.userId = pgPerson.user_id
+    this.salves = pgPerson.salve
+    this.coupons = pgPerson.coupon
+    this.name = pgPerson.name
+    this.email = pgPerson.email
```
craigspaeth: FYI this constructor property assignment could be DRYed up with `Object.assign(this, pgPerson)`

apps/api/models/person/domain.js
```diff
@@ -0,0 +1,114 @@
...
+
+  // We can always add the following methods, depending on what is needed within the backend:
+
+  // getSalves = () => {
+  //   const salves = this.salves
+  //   this.salves = Salve.findForPerson(salves)
+  //   return this
+  // }
+
+  // getCoupons = () => {
```
craigspaeth: These would have to be `async` functions but this could be a nice feature of the OO approach. It _would_ allow us to couple relationships in a slightly nicer to read API...

```javascript
const person = new Person(id)
const coupons = await person.coupons()
```

vs.

```javascript
const coupons = Coupon.findForPerson(id)
```

apps/api/models/person/domain.js
```diff
@@ -0,0 +1,114 @@
...
+        : [],
+    coupons: mergeArrs(
+      person && person.coupons,
+      ((lead.couponCodes && lead.couponCodes.map(id => ({ id }))) || [])
+    )
+  }
+  return _.pickBy(mergedPerson, _.identity)
+}
+
+export const toPG = ({
```
craigspaeth: This should definitely live on the `Person` class and something that might be able to be inherited from other DB backed classes.

apps/api/models/person/resolvers.js
```diff
@@ -0,0 +1,31 @@
...
+const findPersonById = async (_root, args, ctx) => {
+  authorize(ctx)
+  return (await Mapper.findById(args.id)).toGQL()
+}
+
+const updatePerson = async (_root, args, ctx) => {
+  authorize(ctx)
+  const person = await Mapper.findOne({ id: parseInt(args.id) })
+  const personUpdate = Domain.buildPersonUpdate(args, person)
+  await Mapper.update(_.pickBy(personUpdate, _.identity))
```
craigspaeth: These two lines feel like they could go further with the OOP approach. e.g. This could be

```javascript
const person = await Person.findOne({ id: parseInt(args.id) })
await person.save(args)
```

or if we kept the data mapper pattern:

```javascript
const person = await Person.findOne({ id: parseInt(args.id) })
person.update(args)
await PersonMapper.save(person)
```

apps/api/models/person/resolvers.js
```diff
@@ -0,0 +1,31 @@
+import { authorize } from './domain'
+import Person, * as Domain from './domain'
+import * as Mapper from './mapper'
+import _ from 'lodash'
+
+const findPeopleByNameEmail = async (_root, args, ctx) => {
+  authorize(ctx)
+  return (await Mapper.findByNameEmail(args.search)).map(Person => Person.toGQL())
```
craigspaeth: Nitpick, but remember if we go more heavy on the OOP you'll want to be careful about capitalization as that indicates class and lowercase indicates instance—this would be `person.toGQL()`

craigspaeth:
Firstly, I love the organization/grokability wins from simply breaking up modules into domain/resolves/schema/transaction/mapper. I think that alone goes a long way in making the code easier to follow. Later when you compose the pieces inside resolvers the namespaces nicely illustrate how each part comes together.

That said, as I comment about `this`, prototypes vs classes, JSON.stringify, getters/setters, mutating properties, etc. I remember the reasons why I've become scared of OOP 😨 There's a lot of complexity in managing OOP, especially with JS's pitfalls.

I do see how classes/objects could provide a nice central abstraction around a domain and be a way to build a language around relationships e.g. `person.salves().first().sendResultsNotification(practitioner)`. In practice, though that might look more like this when considering async...

```javascript
const practitioner = await Practitioner.findOne(id)
const person = await Person.findOne(id)
const salves = await person.salves()
await salves.first().sendResultsNotification(practitioner)
```

In any case, I think the direction this proposes is not far enough for me to be convinced of adding a lot of OOP into the model layer, as right now it seems like the main benefits illustrated are converting from `pgPerson` to GQL (`(await Mapper.find()).toGQL()` vs `toGQL(await Mapper.find())`—or [methods that wrap functions](https://github.com/findOctave/cortex/pull/854/files#diff-6fda0c30fb37bafaab43231398fa71d3R103) in slightly less verbose names. If you feel strongly about wanting to introduce OOP then I'd encourage spiking a bit further by trying to add an ORM to see if it cleans up some of the transaction/mapper code and/or seeing what it might be like to be more aggressive about replacing the FP stuff with OOP stuff (e.g. removing [intermediate `personUpdate` data](https://github.com/findOctave/cortex/pull/854/files#r284939718) with straight method calls or considering moving the mapper/transaction functions as methods on the Person class to see if you can reuse the fetched state in places).

Also happy to chat more/pair on this IRL. This has definitely been helpful in showing me what introducing mild OOP patterns could be like and while I'm not fully bought-in to what the OOP bit brings to the table, the rest of it looks good and I could be convinced to take the dive. Thanks for taking the time to explore this 🙌

apps/api/models/person/domain.js
lionsdwarf: Good point--thank you for highlighting.

apps/api/models/person/domain.js
lionsdwarf: It works in practice but I too was wondering about unintended side-effects of using *class* here. Thanks for highlighting.

apps/api/models/person/domain.js
lionsdwarf: It's not exactly 1:1 bc of the singular<>plural translation.

apps/api/models/person/domain.js
lionsdwarf: 👍

apps/api/models/person/domain.js
lionsdwarf: 👍

apps/api/models/person/resolvers.js
lionsdwarf: Nice!

apps/api/models/person/resolvers.js
lionsdwarf: Good call!

lionsdwarf:
Thanks for the thorough feedback. Your comments help elucidate where this path would lead should we decide to continue on it. 

At this point--having made some headway on the fp approach--I think our feelings are aligned. The fp spike (which builds off this ooo branch) contains a lot of red. It's more lightweight. I see why you recommend continuing the ooo path with an ORM, since our Mapper is starting to feel like ORM-y already. 

I guess my main concern with the fp approach is the potential for coupling the domain and database layers. Maybe the happy medium is to continue with Mapper sans-ooo (as much as is possible). Not exactly sure how that would look.. but once I PR this next spike, we can circle back and discuss *all the things*!

PR #855, opened 2019-05-17
[ Spike - Do Not Merge ] Fp model breakup

lionsdwarf:
Building off the same file separation implemented in my ooo spike, this approach chips away at the pseudo-ORM Mapper functions by bypassing them in favor of direct knex calls. 

The obvious upside here is the potential elimination of the Mapper functions and the maintenance of that translation layer. On the other hand, eliminating this layer decentralizes a certain degree of query/mutation logic that then gets sprinkled across different parts of the codebase. Also (not to beat a dead horse here),  is the uncomfortable coupling of the data store and the application code, eg column names.

The simplicity in this spike definitely sways me away from the ooo approach, but I'm not 💯 that eschewing the maintenance of an intermediate map will not come without it's own set of headaches and inconsistencies. Maybe joi is the answer. 

Let's discuss the tradeoffs. It may make sense to take both approaches in parallel with different models and see what pans out in each direction. Just an idea...


apps/api/jobs/notifications/assigned-salve.js
```diff
@@ -37,14 +38,14 @@ const sendNotification = async ({
 
 export const sendNotifications = async (people, notification) => {
   const notifications = people.map(async person => {
-    const { salves, leadId } = person
-    const salvesToNotify = getUnnotified(salves)
-    const lead = await at.findOne({ table: 'leads', id: leadId })
+    const { salve, lead_id } = person
```
craigspaeth: I see what you mean here and [here](https://github.com/findOctave/cortex/pull/855/files#diff-9897929069e6e75fd226fb42386be076R27). This definitely does feel awkward to have the downstream data shapes so exposed in jobs code. Even calling functions that imply `pg` or showing underscore casing feels leaky... 🤔

craigspaeth: 🤔  as I ponder on this, I wonder if simply being diligent about pushing as much as possible to the model layer would help mitigate this leakiness.

For instance, in this case, the code could look like...

```javascript
const lead = await Person.findLead(person)
const salvesToNotify = Person.unotifiedSalves(person)
sendNotification({ lead, salvesToNotify, person, ... })
```

...where we're diligent to design functions in a way that are very specific to the question/transaction being made and having a convention of passing full domain objects as the first argument, e.g. `Person.doSomething(person, ...)`. That way we somewhat avoid having to pluck out and manipulate intermediate data between function calls.

We'd also want to be diligent about not exposing terminology that is generic or related to downstream data stores, so instead of `Person.pgUpdate(attrs)` we would expose something a lot more domain-specific like `Person.saveSalvesNotified()`. 

In that case, the body of [the function you call out](https://github.com/findOctave/cortex/pull/855/files#diff-9897929069e6e75fd226fb42386be076R26) would look more like...

```javascript
await Person.saveSalvesNotified(person)
await Lead.sendSalveNotifications(lead, salvesToNotify)
```

That said. I do think there'll be plenty of cases where it's just not really practical to push all the logic down to a single domain model in a way that avoids intermediate references to say snake-cased columns or having to create intermediate maps/arrays like `salvesToNotify`.

In those cases, we could introduce a [service layer](https://martinfowler.com/eaaCatalog/serviceLayer.html) which would be modules that encapsulate a "cross-cutting transactional domain" vs. an "object domain" e.g. a "billing service"  that combines clients/appointments/etc. or "notifications service" that combines salves/leads/content-modules or "authorization service" that co-locates auth transactions across models. This is something I know [larger Rails apps](https://www.engineyard.com/blog/keeping-your-rails-controllers-dry-with-services) and Java-like enterprise stacks do with some success. While I feel our app is still pretty small to get all enterprise/POEAA on it, I also think it might be necessary to introduce earlier given our absence of the magic and architecture that ORMs/more opinionated OOP frameworks provide.

What that would look like in practice is a folder directory like:

```
- apps
  - api
    - services
      - billing
      - notifications
      - etc.
    - jobs
    - lib
    - models
      - lead
      - user
      - coupon
      - etc.
    - test
```

Then code in jobs would be super lean, and this salve notifications job would look something like...

```javascript
export default async () => {
  log.msg('Init notifications/assigned-salve')

  const people = await Person.findWithUnotifiedSalves()
  const notifications = await Notifications.rollupSalveNotifications(people)
  
  if (process.env.IS_NOTIFICATIONS_ACTIVE === 'true') {
    await Notifications.sendSalveNotifications(notifications)
  } else {
    log.msg('Sending notifications', notifications)
  }
  
  log.msg('Fin notifications/assigned-salve')
}
```

Essentially this just pushes a lot of the job code into a different layer, but it might at least help create a home which accomplishes two important things:

1. A place to push code that needs to generate intermediate data or reference granular data like column names to avoid it leaking elsewhere
2. A way to separate any code that doesn't directly have to do with the interaction between "domain models" or "business logic" (e.g. like how the code above leaves the logging and environment config outside of the services layer).

Phew! Let me know what you think!

craigspaeth:
Excited we have a direction to go now! 🙌 👏 Good stuff!! Going to close this spike PR, but excited to see what comes next!

PR #856, opened 2019-05-20
Cache rules everything around me... make profile page query faster

craigspaeth:
This takes the profile page query down from several seconds to ~500ms by liberally caching on the client + appointment + practitioner/type related queries (+ less importantly the salve + salve categories).

😭 😭 😭 😭 😭 

For now, this should hopefully mitigate some performance problems with the profile page. It's still slow on initial load via a cache-miss, but I'm hopeful that some of the sub-queries like an appt's practitioners/types can be reused making this acceptable for now. Later we'll want to look into either warming up the cache on load by requesting these on server boot-up... or more likely we'll want to stop playing whack-a-mole and TLDR; Build Your Own EHR and/or build some kind of continuous snapshotting system like I introduced a while back.


craigspaeth:
Gonna merge b/c this is all just config and I need to see it live to test it on a running staging/prod system.

PR #858, opened 2019-05-20
[ WIP -- Do not merge ] Fp model breakup

lionsdwarf:
Applies the "service" layer across our `/assign` app and `assigned-salves` job. Putting up this WIP to get a second pair of eyes on the initial implementation. 

If anything here looks suspicious/leaky, please highlight as I am currently applying this pattern across our other `Person` references.


apps/api/jobs/notifications/assigned-salve.js
```diff
@@ -1,63 +1,18 @@
...
   log.msg('Init notifications/assigned-salve')
-  const notification = (await at.find({
-    table: 'notifications',
-    filter: `name = 'assignedSalve'`
-  }))[0]
-  const peopleToNotify = await Person.jobs.findBySalvesToNotify()
-  await sendNotifications(peopleToNotify, notification)
+  const notifications = await Notifications.salves()
+  if (process.env.IS_NOTIFICATIONS_ACTIVE === 'true') {
+    await Notifications.send(notifications)
```
craigspaeth: Nice! I'm liking the cleanup a service module provides here.

apps/api/models/lead.js
```diff
@@ -76,8 +76,8 @@ export const fromAirtable = record => ({
 })
 
 const joinSalves = async lead => {
-  const { salves } = await Person.jobs.findOne({
-    leadId: lead.id
+  const { salves } = await Person.findWhere({
```
craigspaeth: Nit-pick: In light of the "more domain specific/passing around full objects" this could be `Person.findByLead(lead)`

apps/api/models/salve.js
```diff
@@ -151,7 +156,7 @@ export const findById = async salveIds =>
     : []
 
 export const findAllForClient = async patientId => {
-  const person = await Person.jobs.findOne({ patientId })
+  const person = await Person.findOne({ patientId })
```
craigspaeth: 👍 liking the removal of nested namespace now

apps/api/services/notifications.js
```diff
@@ -0,0 +1,43 @@
+import { sendEmail, sendSMS } from '../lib/notify'
+import * as Lead from '../models/lead'
+import * as log from 'log'
+import * as at from 'at'
+import Person from '../models/person'
+
+const { APP_URL } = process.env
+
+export const getUnnotified = salves =>
```
craigspaeth: I know I wrote this but this could probably live inside the salve model (domain) seeing as it's logic localized to that domain.

apps/api/services/notifications.js
```diff
@@ -0,0 +1,43 @@
...
+  }))[0]
+  const people = await Person.findByUnnotifiedSalves()
+  const notifications = await Promise.all(people.map(async person => {
+    const lead = await at.findOne({ table: 'leads', id: person.lead_id })
+    return {
+      lead,
+      person,
+    }
+  }))
+  notifications.content = content
```
craigspaeth: 👍 I like how this service concepts gives some room to create intermediate models and this function in particular seems like a good example of its utility. It's logic that purely crosses multiple models and combines them. One nitpick is `Notifications.salves` is a bit ambigious, maybe something like `Notificaitons.rollupUnotifiedPeopleAndContent` or something terser than that. `salves` is just a bit confusing b/c this funcitons doesn't return salve data of any sort.

craigspaeth:
👏 This is awesome! I think this is very promising for how this direction can help clean things up. Looking over the mapper/transactions and resolves/schema/toGQL relationship I'm wondering if it'd make sense to consolidate these things (sry I keep going back and forth on how to slice things 🙃).

Could we instead start by breaking down models into 3 parts?
- **gql.js** Contains resolvers, `toGQL` functions, and type defs GQL schemas
- **domain.js** Just like it is now, grab bag of "business logic", probably mostly pure functions
- **mapper.js** Combination of mapper and transaction. Any logic that has to do with talking to the data-store. Probably exposes more domain-sounding functions like `Person.findByLead` and `Person.applyCoupon` than the "transaction" functions like `findWhere` or `upsert`

Then of-course the services pattern remains to put those pieces together. This might help use start out with less layers/overhead. 

In any case, that and my comments above are pretty much nitpicking. This is great and I'm excited to merge this when you're ready and apply this across the board! 👏

apps/api/jobs/notifications/assigned-salve.js
lionsdwarf: 👍

apps/api/services/notifications.js
lionsdwarf: I might argue that, by virtue of being a property on the `Notifications` domain, `salves` in this context is akin to a "salve notification", but hear what you're saying. I think this debate reveals the differences in our OO vs FP tendencies. I am not strongly opinionated on it.

craigspaeth: 👍 you make a good point, reading it all together that makes sense. Not strongly opinionated either. Leave as is!

PR #864, opened 2019-05-23
Config to less chatty cache settings and load profile appt data on client-side

craigspaeth:
I realized there's a good chance that some pages like the billing page are having trouble loading because of my liberal sprinkling of `refreshAhead` caching for the profile page to load faster. This could cause a ton of background intervals to run and try to keep the cache warm, keeping Airtable's API busy.

I also noticed this was super hard to repro locally, and this might be because [Airtable throttles based on IP](https://community.airtable.com/t/is-the-api-rate-limit-per-api-key-or-app-id/2265). So prod. could have over time accumulated a lot of refreshAhead intervals and Airtable could have been rate limiting cache-miss requests (like the request for a lead on the billing page). Supporting this theory is that aptible restart seemed to help the problem (causing the background intervals to get wiped).

This PR tones that down by only keeping `refreshAhead` where it truly needs to be, increasing TTLs, removing caching from appointments, and deferring the appointment rendering to the client-side to keep the initial page load fast (to my dismay as I hate solving performance problems by deferring it to the client-side, but just trying to get things stable first before exploring other options).

Mo' cache... mo' problems. 😭


apps/portal/lib/view-model.js
```diff
@@ -116,6 +116,55 @@ export const newAppointment = state =>
     _.includes(['THERAPY', 'COACHING'], appt.type.category)
   )
 
+export const apptsQuery = (state, now = moment()) => {
+  const startOfToday = moment(doesNotThrow)
+    .utc()
+    .startOf('day')
+    .format()
+  return `
+    fragment ClientAppointment on Appointment {
```
craigspaeth: These are moments that makes me love isomorphic JS. To defer the appt. rendering to the client-side was a relatively simple cut and paste of this query from `getInitialProps` to the view model.

If this were a different polyglot stack like a traditional Rails stack one would need to do a ton of work to support JSONifying Rails models, exposing that JSON in controllers, rewrite the controller logic to fetch that JSON data, and finally rewrite the ERB view into React components or some jQuery spaghetti 😵

lionsdwarf: jQuery spaghetti...yummmm!

Win!

lionsdwarf:
Curious why the appointments query had to move client side but LGTM!

craigspaeth:
I moved the appts. query client-side because it took at-least half the load time on the page (TLDR Build yo blah blah).

Moving it client-side is a smoke and mirrors performance trick to make it so rendering happens staggered and people are able to at-least access billing and salve features while appointments load in parallel.

lionsdwarf:
Copy that!! Thanks for the clarification.

PR #872, opened 2019-05-24
Sort salves by title in admin

craigspaeth:
Small request by Murk to alpha-sort salves for easier scanning—didn't even feel worth a test how small this was.


craigspaeth:
Mergin' deployin'

PR #875, opened 2019-05-29
[ WIP -- DO NOT MERGE ] Domain/Mapper pattern for models, Kill Leads table

lionsdwarf:
Per the #help channel, looks like I may be shifting gears into bug-squashing mode, so wanted to put this WIP up. No urgency here whatsoever.

This is still very broken/incomplete, but should give you a general direction of where things are heading at the moment.

As discussed on Slack....This refactor requires a lot of re-writing code around the Lead model, which doesn't really make sense since we are killing the Leads table. So after restructuring the Person model according to our Domain/Mapper pattern strategy, I traced all read/writes to Leads table and have started moving those transactions onto the Person table as part of the model refactor. 

Since you are traveling, I suggest you pick this up NEXT WEEK!


lionsdwarf:
Some good patterns starting to emerge here... 

Queries and mutations for `Person` have been refined. As recovering RESTafarians, we implemented some gql anti-patterns with, for instance, multiple person queries, e.g. `findPersonById`, `findPersonByEmail` and had gone down a similar path w mutations (I think some other examples exist in different models too ☹️ ). 

Instead, we now have single 'find', 'update', 'upsert' resolvers that leverage gql [query arguments](https://www.prisma.io/blog/designing-powerful-apis-with-graphql-query-parameters-8c44a04658a9). This intuitively simplifies the api and data fetching, and it will also allow us to more easily extend these resolvers in non-breaking ways.

Also in positive news: the Lead model is almost entirely gone.

PR #878, opened 2019-05-30
Revert "Fix for Reschedule/Schedule appointment in Portal app"

lionsdwarf:
Reverts findOctave/cortex#876

Schedule consult is not working in staging even though works in prod and on local master. Error Error throwing from a dependency it seems, as we do not use `.split` anywhere in `onboarding` app.

Reverting to test state prior to merging fix code. Not sure what else to do since I cannot repro in any other environment...


lionsdwarf:
This did not fix the issue 🤔

PR #882, opened 2019-06-05
Render errors on profile scheduler submit button

craigspaeth:
This addresses one of Sin's clients issues with trying to reschedule an appt. Digging into Apollo Engine's error traces I noticed some "Appointment already booked" errors surfacing around the reported time. It appears this person was probably hitting these b/c of cache staleness 😭 and seeing the button just hang.

I dug into it and realized that we weren't rendering errors from the submission. So this adds that and alongside it I quietly cry inside knowing the better fix would be no caching... but then the page would load too slow... and so TLDR; Build your own EHR.


lionsdwarf:
🤔 Hmm Interesting. This functionality worked previously because I remember seeing this error message in the past. In any case, LGTM! Thanks!

PR #883, opened 2019-06-06
Add outdated browser notice

craigspaeth:
Uses this [outdated-browser-rework](https://github.com/mikemaccana/outdated-browser-rework) module to pop up a banner that asks people with old browsers to upgrade.


lib/components/outdated-browser/index.css
```diff
@@ -0,0 +1,72 @@
+:global #outdated * {
```
craigspaeth: The module uses SASS, so I just copy-pasted the compiled styles into a CSS module here.

lionsdwarf:
Easy enough! Thanks!

PR #885, opened 2019-06-06
Kill env set scripts

craigspaeth:
Correct me if I'm wrong but it seems like Aptible actually has a [rebuild](https://www.aptible.com/documentation/enclave/cli/rebuild.html) script unlike Heroku so we don't really need to have this wrapper anymore. It should be as simple as:

```
aptible config:set FOO=BAR --app staging
aptible rebuild --app staging
```


lionsdwarf:
Good find. Wasn't aware of this. 👍

PR #889, opened 2019-06-10
Update dependencies and node version

craigspaeth:
This updates our dependencies and node version to the latest including Next.js 🎉 

Regarding the Next upgrade—I found the [problematic piece of state in next-css](https://github.com/zeit/next-plugins/issues/373#issuecomment-500260362) that was causing the type error and was able to work-around that issue by checking in a modified version of next-css (kinda hard to simply fork the mono-repo they're using). Hopefully, in the not too distant future, the Next team & I get a fix merged in master and we can go back to using the core next-css module.

The other upgrades required a lot of finagling of webpack/babel/next config files but was otherwise pretty straightforward.

Let's hold off on merging this until after we deploy your big PR—I don't want two huge upgrades going out at once in case we need to fire-fight one or the other.


lionsdwarf:
Impressive sleuthing! Will hold off on merging for now. Good work! 👍

PR #904, opened 2019-06-13
Pulsing loading-state placeholder for Appointments on Profile page

lionsdwarf:
appt_placeholder


apps/portal/components/appointments/index.css
```diff
@@ -119,3 +119,76 @@
...
+    opacity: 0.7;
+  }
+  50% {
+    opacity: 0.6;
+  }
+  55% {
+    opacity: 0.7;
+  }
+  60% {
+    opacity: 0.8;
```
craigspaeth: Quite a specific tweening going on here! FWIW you might be able to use [css easing](https://matthewlein.com/tools/ceaser) in cases like this.

apps/portal/components/appointments/index.js
```diff
@@ -13,6 +13,23 @@ const cx = classNames.bind(styles)
 const isWithin48Hrs = startTime =>
   moment(startTime).isBefore(moment().add(2, 'days'))
 
+const ApptPlaceholder = () => (
```
craigspaeth: 👍 Nice extracting of component here.

craigspaeth:
woot—good stuff thanks!

apps/portal/components/appointments/index.css
lionsdwarf: The specificity is to create a pause between pulses. Constant pulsing felt overly busy/distracting. Would easing manage that ? My stackoverflowing on animation pausing revealed nothing related to easing.

craigspaeth: Ah right right—nope good call. There is `animation-delay` which would cause pauses between loops, but I'm all for the %s. 👍 good stuff!

lionsdwarf:
I believe animation delay will only pause start of animation

PR #907, opened 2019-06-19
Sync Cliniko users > Auth0 users

lionsdwarf:
Let's deploy this, *then* we'll apply the roles manually as discussed IRL.


craigspaeth:
👍 easy peasy.

PR #914, opened 2019-06-24
Pull practitioner from the latest sessions and model refactors

craigspaeth:
This is mostly a lot of _Scout's Rule_-ing by refactoring the appointment and practitioner model into the new architecture. I'll comment on the part that is new functionality.


apps/api/models/practitioner/mapper.js
```diff
@@ -0,0 +1,58 @@
...
+    appts.map(appt =>
+      cliniko.find({
+        resource: 'appointment_types',
+        url: appt.appointment_type.links.self
+      })
+    )
+  )
+  const practitioner = await cliniko.findOne({
+    resource: 'practitioners',
+    url: Appointment.latestSession(appts, apptTypes).practitioner.links.self
```
craigspaeth: This is the new bit of functionality that solves the TODO and finds the latest therapy/coaching session to pull the practitioner from. I think it was hard to reproduce if you were using a patient that had existing appointments because Cliniko sorts them by `created_at` ascending—meaning it would always return the first appointment in the patient's history even as you try to add new appointments.

lionsdwarf: Ah, `created_at`, got it....

PR #915, opened 2019-06-24
set billing page button active

lionsdwarf:


craigspaeth:
Would love to see a unit test with this, but I’ll let that be a follow up PR

PR #922, opened 2019-06-26
Dynamic business id + location model

lionsdwarf:


apps/api/models/appointment/mapper.js
```diff
@@ -31,22 +30,24 @@ export const findPractitionerUser = async appt => {
...
-export const create = ({
+export const create = async ({
   patientId,
   startTime,
   appointmentTypeId,
   practitionerId
-}) =>
-  cliniko.create({
+}) => {
+  const location = await Location.findByPractitioner({ id: practitionerId })
```
craigspaeth: 👍 Nice!

apps/api/models/location/mapper.js
```diff
@@ -0,0 +1,30 @@
...
+  let locations = await findAll()
+  locations = await Promise.all(
+    locations.map(async location => {
+      location.practitioners = await Practitioner.findAllForLocation(location)
+      return location
+    })
+  )
+  return _.find(
+    locations,
+    _.flow(
```
craigspaeth: OOo look at you FP wizard using `_.flow` 👍

apps/api/models/location/mapper.js
```diff
@@ -0,0 +1,30 @@
...
+import * as cliniko from 'cliniko'
+import _ from 'lodash'
+
+const findAll = () =>
+  cliniko.find({
+    resource: 'businesses',
+    all: true
+  })
+
+export const findByPractitioner = async practitioner => {
```
craigspaeth: I think this could just be `(practitionerId) =>` seeing as we only use the ID in this function and the one above isn't sending a full practitioner object.

apps/api/models/location/mapper.js
```diff
@@ -0,0 +1,30 @@
...
+  let locations = await findAll()
+  locations = await Promise.all(
+    locations.map(async location => {
+      location.practitioners = await Practitioner.findAllForLocation(location)
+      return location
+    })
+  )
+  return _.find(
```
craigspaeth: This last `_.find(..)` part could probably be extracted into a pure function with tests.

craigspaeth:
Looking good. These two comments would be ideal to address but after that feel free to merge when 💚 
- https://github.com/findOctave/cortex/pull/922/files#r297896517
- https://github.com/findOctave/cortex/pull/922/files#r297896684

craigspaeth:
👏 Beauty! Well done.

PR #926, opened 2019-06-28
CSS changes

UroojNawaz:
I made the images bigger


apps/marketing/components/Post/post.css
```diff
@@ -1,7 +1,7 @@
 @import 'styles';
 
 .root {
-  max-width: 680px;
+  max-width: var(--columns-wide-8);
```
lionsdwarf: Nice work using the css variables!

PR #927, opened 2019-07-02
Appointment Notifications

lionsdwarf:
Appointment notifications for: 
1) Individual Appointments and Group Attendees confirmed within the last hour 
2) Individual Appointments and Group Attendees with Appointment `starts_at` within the upcoming 24 hours.

Two things currently needed to make this prime-time ready:
1) Timezone datetime coercion 
2) Worker config


apps/api/services/notifications.js
```diff
@@ -1,17 +1,51 @@
...
+    })
+  )
+  apptRollups.content = content
+  return apptRollups
+}
+
+export const sendAppointmentNotification = apptRollups =>
+  Promise.all(
+    apptRollups.map(appt => {
+      return sendEmail(appt.recipient, apptRollups.content, {
```
craigspaeth: Nit-pick: You can leave out the `return` and curlies here to win at code ⛳️

apps/api/services/notifications.js
```diff
@@ -1,17 +1,51 @@
...
-    filter: `name = 'assignedSalve'`
+    filter: `name = '${name}'`
   }))[0]
+
+export const upcomingAppointments = async () =>
+  appointmentRollups(await Appointment.findStartsBetween(), 'remindAppointment')
+
+export const confirmedAppointments = async () =>
+  appointmentRollups(
+    await Appointment.findCreatedWithin(),
```
craigspaeth: Nit-pick: The defaults actually make this kind hard to read—might be better to leave it exposed 🤷‍♂

craigspaeth:
👏 really good stuff here! This is an excellent starting point. Also one more item for the prime-time list: tests.

apps/api/services/notifications.js
lionsdwarf: Good catch. Thxxx

apps/api/services/notifications.js
lionsdwarf: What do you mean by exposed exactly?

lionsdwarf:
Closing this for now. Thanks for the feedback. Will definitely add tests.

apps/api/services/notifications.js
```diff
@@ -1,17 +1,51 @@
...
+    'confirmAppointment'
+  )
+
+export const appointmentRollups = async (appts, contentName) => {
+  const content = await findContentByName(contentName)
+  const apptRollups = await Promise.all(
+    appts.map(async appt => {
+      appt.recipient = await Patient.findForAppointment(appt)
```
craigspaeth: Noticed an issue here. This assumes a 1:1 patients to appointment—meaning it won't work for groups where there are many:1.

apps/api/services/notifications.js
craigspaeth: Just removing the defaults so it reads `Appointment.findCreatedWithin(moment(), moment().add(24, 'hours')` which reads a bit more explicitly to me. (like I said—just a nit pick)

PR #933, opened 2019-07-04
Custom appointment reminders + confirmations

craigspaeth:
This piggy-backs on your last PR to include:

- Scoped notifications by appointment type
- Inject email vars like `patientPreferredFirstName` and `appointmentTypeName`
- Adjust the timezone based on the business zip code
- Attach an ics file

image


apps/api/models/appointment/mapper.js
```diff
@@ -74,3 +74,29 @@ export const findGroupAppointment = async appointmentId =>
...
   }))[0]
+
+export const findBookingsCreatedSince = async time => {
+  const attendees = await cliniko.find({
+    resource: 'attendees',
+    q: `created_at:>${moment.utc(time).format()}`
+  })
+  return Promise.all(
+    attendees.map(attendee =>
+      cliniko.find({ url: attendee.booking.links.self })
```
craigspaeth: Cliniko calls something that's either an individual or group appointment a "booking", so I'm refactoring the naming here to follow to that. It also reflects the data model being returned here more directly.

I also refactored the coercing of group appointments into individual appointments because I found it confusing to be handling an `appt` object but not knowing if it was truly raw appointment data, something returned from bookings, or one of the coerced objects.

apps/api/models/content-module.js
```diff
@@ -199,3 +199,14 @@ export const queries = {
   contentModule,
   contentModules
 }
+
+export const contentFormula = (name, appointmentTypeId) =>
+  appointmentTypeId
+    ? `AND(name='${name}', FIND('${appointmentTypeId}', appointmentTypes))`
+    : `name=${name}`
+
+export const find = async (name, appointmentTypeId) =>
```
craigspaeth: Moved the `findContent` function out to the content-module function as it was only dealing with this model and not the "rollups" or combination of models.

apps/api/models/appointment/mapper.js
lionsdwarf: Nice cleanup! I admit to getting a little sloppy with my domains there. Will be more conscientious of that going fwd.

apps/api/models/patient.js
```diff
@@ -107,3 +107,15 @@ export const addReferralNote = async (patient, referralNote) => {
     data: referralToTreatmentNote({ patient, provider, referralNote })
   })
 }
+
+const findFrom = relationship =>
+  cliniko.findOne({ url: relationship.patient.links.self })
```
lionsdwarf: Nice. I'd been struggling with naming these types of functions. I like how the generic argument name to make this reusable.

apps/api/models/practitioner/mapper.js
```diff
@@ -63,3 +74,9 @@ export const findAllForLocation = location =>
     resource: 'practitioners',
     url: location.practitioners.links.self
   })
+
+export const findFrom = relationship =>
+  cliniko.find({
+    resource: 'practitioners',
```
lionsdwarf: Would want to confirm w a test, but I believe the url property makes this `resource` prop superfluous.

craigspaeth: It does work, but I'm doing this maybe overly-clever thing where it tries to un-nest the namespace so one doesn't have to do `(await cliniko.find(...).practitioners)` https://github.com/findOctave/cortex/blob/master/lib/cliniko/index.js#L99

craigspaeth:
For sure! You did such a good job setting me up to wrap the last pieces of this 🙌 Teamwork is the dreamwork 🌈 ✨

PR #941, opened 2019-07-08
[WIP - Do not merge] Kareo poc

lionsdwarf:
Consumes the Kareo SOAP api using the approach detailed here: https://stackoverflow.com/a/22897376/5656957


lionsdwarf:
This currently makes a "getProviders" callout and `console.log`s the response.

craigspaeth:
👏 Well done! Slogging through SOAP/XML I imagine is no fun task.

The code looks great so far, I only have a couple of high-level thoughts relevant to what we might think about in the future of growing this code:
- Having all of this code live in "services" feels off, seeing as there's very little cross-model-coordinating going on here right now. (I'd like to keep services pretty strictly about cross-model/domain functionality and avoid services just becoming a differently named lib folder.) 
- It'd be great to abstract out a kareo "lib" that takes care of the low-level parts by accepting arguments that build the unique parts of the XML and encapsulates the initializing w/ env vars, common templating logic, and response handling.
- After that, I think we can consider using that kareo lib inside relevant models. (e.g. `Client.submitInsuranceInfo()` or `Practitioner.syncInsuranceInfo()`).

lionsdwarf:
Hear ya loud-n-clear, Craig, and on the same page....Closing for now.

PR #945, opened 2019-07-09
Accommodate timezone based on zip code in consult flow

craigspaeth:
This picks different content modules for the billing screen based on if the zip-code is in California. That way we can communicate to CA-based clients that they can use our insurance partners.

image

It also renders timezone in the scheduler and determines based on zipcode or guessed from IP address (using `moment().tz().guess()`).

Screen Shot 2019-07-09 at 6 37 38 PM


apps/onboarding/lib/view-model.js
```diff
@@ -69,6 +70,94 @@ export const initialState = {
   errorMsg: null
 }
 
+export const consultContentQuery = ({
```
craigspaeth: This is a _Scouts Rule_ thing where I moved the duped consult queries b/t class and regular consults into the view model.

apps/onboarding/lib/view-model.js
```diff
@@ -433,6 +522,11 @@ export const nextStep = state => ({
   errorMsg: null
 })
 
+export const billingContent = (state, content) =>
+  zipcodes.lookup(state.person.zipCode).state === 'CA'
```
craigspaeth: This is the bit that uses [zipcodes](https://www.npmjs.com/package/zipcodes) to determine if they're based in CA and pick a different content module if so.

lib/availability-picker/index.js
```diff
@@ -114,6 +123,7 @@ const Availabilities = ({
 )
 
 export default ({
+  timezone = moment.tz.guess(),
```
craigspaeth: Here we accept passing in the timezone and default to guessing based on the computer's IP address.

lib/availability-picker/index.js
```diff
@@ -56,30 +58,35 @@ const AvailabilitiesList = ({
...
-                formatDatetime(selectedAvailability.startTime)
-              ? 'is-selected'
-              : ''
-          )}
-        >
-          {moment(availability.startTime).format('h:mm a')}
-        </div>
-      ))}
+      {availabilities.map(availability => {
+        const tzStartTime = momentTz(availability.startTime).tz(timezone)
```
craigspaeth: Here we adjust the availabilities to the timezone.

apps/onboarding/components/schedule-consult.js
```diff
@@ -68,6 +69,7 @@ export default props => {
       ),
       body: (
         <AppointmentForm
+          timezone={zipToTz.lookup(state.person.zipCode)}
```
craigspaeth: Here we pass in the timezone based on the entered zip code. We're also using this library in appointment reminder/confirmations—isomorphic JS FTW!

apps/onboarding/lib/view-model.js
```diff
@@ -517,6 +613,16 @@ const trackStepAndScrollUp = state => {
   window.scrollTo({ top: 0 })
 }
 
+export const confirmParagraph = (state, content) => {
```
craigspaeth: This injects the "See you {{onDate}}" variable but now corrected to the client's timezone and including the TZ code.

apps/onboarding/components/schedule-consult.js
lionsdwarf: Rock on!! 🎸

lib/availability-picker/index.js
lionsdwarf: 👍

lib/availability-picker/index.js
lionsdwarf: Nice method

PR #946, opened 2019-07-10
Config Sentry to work server-side

craigspaeth:
Sentry stopped supporting Raven and I think that may have resulted in us losing server-side errors being reported in Sentry. This should fix that, and hopefully, help debug issues like the 500 error Sahar's client encountered.


craigspaeth:
Just some config change, so gonna merge to see if we get reporting in staging/prod.

PR #949, opened 2019-07-10
Pick California based intakers 

craigspaeth:
Ops will have a different intake coordinator for SF. Luckily the timezone preferences are exposed in Cliniko's API. This filters down to the California-based intaker's availabilities if the client is located in Cali or across all non-Cali intakers if not Cali (some-day we might be lucky enough to directly match timezone-to-intaker when Octave expands across _all the timezones_).


apps/api/models/practitioner/mapper.js
```diff
@@ -54,20 +54,17 @@ export const findForPatient = async patientId => {
...
-export const findUserForPatient = async patientId => {
-  const practitioner = await findForPatient(patientId)
-  const user = await cliniko.findOne({
-    resource: 'users',
-    url: practitioner.user.links.self
-  })
-  return user
-}
+export const findUserForPatient = async patientId =>
+  findUser(await findForPatient(patientId))
```
lionsdwarf: 👍

PR #950, opened 2019-07-10
[Do not merge] Kareo lib with get and create methods

lionsdwarf:
kareo_create
kareo_get


lib/node-kareo/soap-envelope.js
```diff
@@ -0,0 +1,58 @@
...
+export const requestHeader = `<sch:RequestHeader>
+  <sch:ClientVersion>2.1</sch:ClientVersion>
+  <sch:CustomerKey>${KAREO_CUSTOMER_KEY}</sch:CustomerKey>
+  <sch:Password>${KAREO_IT_PASSWORD}</sch:Password>
+  <sch:User>${KAREO_API_USER_EMAIL}</sch:User>
+</sch:RequestHeader>`
+
+export const buildObj = (soapMeta, payload) => {
+  return `<sch:${soapMeta.object}>
+    <sch:FirstName>${payload.FirstName}</sch:FirstName>
```
lionsdwarf: These fields are still hard-coded, but will be dynamic a la https://github.com/findOctave/cortex/pull/950/files#diff-b9f25232d8ea6bdfc138aa62f2074381R37

lib/node-kareo/test/soap-envelope.test.js
```diff
@@ -0,0 +1,8 @@
+import * as SE from '../soap-envelope'
+
+describe('buildFieldsList', () => {
+  test('builds the fields list', () => {
+    const fl = SE.buildFieldsList(['FirstName', 'LastName'])
+    expect(fl).toEqual('<sch:FirstName>true</sch:FirstName><sch:LastName>true</sch:LastName>')
+  })
+})
```
lionsdwarf: Lots more tests to come....

lib/node-kareo/index.js
```diff
@@ -0,0 +1,25 @@
...
+const req = async (soapObject, params, action) => {
+  const soapMeta = buildSoapMeta(soapObject, action)
+  const soapEnv = SoapEnvelope[action](soapMeta, params)
+  const res = await request(soapEnv, soapMeta.actionObject)
+  return response({ res, soapMeta })
+}
+
+export default ({
+  create: (soapObject, params) => req(soapObject, params, 'Create'),
+  get: (soapObject, params) => req(soapObject, params, 'Get'),
```
craigspaeth: As discussed on the call we may want to be careful about assuming too much about the various "Create" and "Get" calls _actually_ being consistent enough to write abstractions over. It does seem like there's probably an opportunity to reuse code around most of them, but we may want to keep that an internal implementation detail and expose much more explicit APIs with this library.

For instance, maybe it makes more sense to expose a function for each individual available action like...

```javascript
kareo.getPatient(...)
kareo.createEncounter(...)
```

...and most of those can reuse the same params/response handling functions internally, but we try to expose a minimalistic/explicit function signature for each.

lib/node-kareo/response.js
```diff
@@ -0,0 +1,33 @@
...
+    case 'Patient':
+      return {
+        id: body['PatientID'][0]
+      }
+  }
+}
+
+export default ({ res, soapMeta }) => new Promise(async (resolve, reject) => {
+  const jsRes = await parseRes(res)
+  const resBody = jsRes['s:Envelope']['s:Body'][0][`${soapMeta.actionObject}Response`][0][`${soapMeta.actionObject}Result`][0]
```
craigspaeth: Gnarly responses! Nice work slogging through this.

craigspaeth:
👏 good stuff! This is a great start and will make interacting with Kareo so much nicer. I look forward to the day this is robust, we have an engineering blog, and you can release this beauty into the open source world!

lib/node-kareo/index.js
lionsdwarf: 👍

PR #954, opened 2019-07-11
Fixes on confirm copy

craigspaeth:


apps/onboarding/lib/view-model.js
```diff
@@ -628,10 +628,12 @@ const trackStepAndScrollUp = state => {
 }
 
 export const confirmParagraph = (state, content) => {
+  if (!state.appointment) return content.p[0]
   const tzApptTime = momentTz(state.appointment.startTime).tz(
     zipToTz.lookup(state.person.zipCode)
   )
-  const onDate = `${tzApptTime.calendar()} ${tzApptTime.format('z')}`
+  const calTime = tzApptTime.calendar(null, { sameElse: 'L h:mm A' })
```
craigspaeth: `moment.calendar()` does this smart human readable thing where it displays a time relative to the current time e.g. "tomorrow at 8:30 am" or "Sunday at 8:30 am". For some reason it drops the time when it's over a week later, so this `sameElse: {}` bit formats that back to include the time.

lionsdwarf: Ah interesting. What an amazing feature!!! Thank you moment!!

PR #956, opened 2019-07-11
Use correct bookings query for confirmations

craigspaeth:
We were querying attendees by created_at data. The problem with `created_at` date is it's based on when the record was created in the database—not related to the appointment data. This replaces that with `Appointment.findBookingsBetween` which works for both use cases.


craigspaeth:
Going to combine this with another PR

PR #959, opened 2019-07-11
Revert findBookings "fix"

craigspaeth:
D'oh confirmations _are_ based on created_at b/c they should happen immediately after the appt. is created _not_ after the appt's start time. Dunno what I was thinking when I was testing and came up with this https://github.com/findOctave/cortex/pull/956


lionsdwarf:
I'm remiss as well. I should have given it closer consideration. My apologies!

PR #966, opened 2019-07-16
Specify single next app to mount via cmd line args

lionsdwarf:
Going to merge so that Urooj can move fwd....


lionsdwarf:
The test is failing on master branch so I'm merging

craigspaeth:
👍 good stuff, thanks!

PR #971, opened 2019-07-17
Refactor onboarding view model into separate parts

craigspaeth:
This pretty simply splits up the view model functions in the onboarding library into separate related parts...

- billing
- class
- consult
- transfer
- copy
- person
- util


craigspaeth:
Still working through some updates it seems

PR #973, opened 2019-07-17
Fix billing page race condition

craigspaeth:
Adds a setTimeout to force the view model reducers to run after next tick/render cycle.


craigspaeth:
Going to merge these two PRs to get the codebase in a clean place for Tanj.

PR #974, opened 2019-07-18
Fix appointment error bug and add a test for it

TMcMeans:
Pushing my first PR! Yay!


apps/portal/test/lib/view-model.test.js
```diff
@@ -178,5 +178,15 @@ test('addAppts adds loaded appointments', () => {
     error: null,
     stripe: null,
     submitting: false
-  })
-})
+  });
+});
+
+test('addAppts changes loading state', () => {
```
craigspaeth: 👏 Test on day 1!?!?!?!

craigspaeth:
Awesome! Just need to quickly run the linter.

PR #976, opened 2019-07-19
Replace "upsert person" with "create if not exists"

lionsdwarf:
Lotsa moving pieces on this still but it will come out the other side with some nice design and debt improvements. Will add comments inline.


apps/api/models/client.js
```diff
@@ -66,13 +74,28 @@ export const schema = {
...
+    ])
+    await Person.update(
+      {
+        userId: user.user_id,
+        customerId: customer.id
+      },
+      { id: person.id }
+    )
+  }
+  return person.id
```
lionsdwarf: Previously this function was only concerned w/ upserting Customer and User, then updating the respective ids on the Person. This approach relied on an assumption that the Person record already existed, which in turn required the frontend to manually call a create-person mutation beforehand. 

This new approach obviates the prerequisite create-person mutation by handling the person record insert, if necessary.

The intertwined-ness of the Person/Client domains here makes me feel even more encouraged that we should really be treating the Client as a child of the Person, instead of as a sibling.

Accordingly, I think my next step is to move this mutation onto the Person record, which I think will be a good first boy-scout step toward resolving a little bit of data mis-modeling debt.

apps/api/models/person/domain.js
```diff
@@ -17,21 +17,6 @@ export const authorize = (args, ctx, ast) => {
...
-    patientId: (patient && patient.id) || (person && person.patientId),
-    customerId: (customer && customer.id) || (person && person.customerId) || (user && user.user_metadata && user.user_metadata.customerId),
-    email: (patient && patient.email) || (person && person.email),
-    name: (patient && patient.name) || (person && person.name),
-    salves: (person && person.salves) || [],
-    coupons: (person && person.coupons) || []
-  }
-  return _.pickBy(mergedPerson, _.identity)
-}
-
```
lionsdwarf: Super syked to be killing off this tumor of a function! It's been taunting me for some time now.

apps/api/models/person/gql.js
```diff
@@ -104,16 +100,16 @@ const updatePerson = async (_root, args, ctx, ast) => {
...
-    },
-    args.by
-  )
+  let [person] = await findWhere({ email: args.email })
+  if (person) throw new Error('Account already exists')
+  const patient = await Patient.upsert(args, 'email')
+  person = (await insert({
+    ...args,
+    patientId: patient.id
+  }))[0]
```
lionsdwarf: Here's the core of the functionality outlined in the ticket. Most of the changes in the PR are side-effects of this change.

apps/onboarding/lib/view-model/billing.js
```diff
@@ -4,22 +4,31 @@ import zipcodes from 'zipcodes'
...
-    mutation {
+    mutation CreateClient($input: PersonClientInput!){
       upsertPersonClient(
-        personId: "${state.person.id}"
-        stripeSource: "${token.id}"
+        input: $input
       )
     }
-    `
+    `, { input }
```
lionsdwarf: Frontend using new "input" payload for mutations. I think we should stick to this approach mainly because we can avoid cumbersome template strings and just pass a regular js object instead.

apps/api/models/client.js
```diff
@@ -41,11 +41,19 @@ export const schema = {
...
+      zipCode: Int
+      birthDateTime: String
+      genderPronouns: String
+    }
   `,
   mutations: `
     upsertPersonClient(
-      personId: String!
-      stripeSource: String!
+      input: PersonClientInput!
```
lionsdwarf: Mutation input type. More info here: https://blog.apollographql.com/designing-graphql-mutations-e09de826ed97 and here: https://graphql.org/graphql-js/mutations-and-input-types/

apps/api/models/person/gql.js
```diff
@@ -48,9 +45,8 @@ export const schema = {
       couponIds: [ID]
       couponCodes: [String]
     ): Person
-    upsertPerson(
-      by: PersonUpsertColumn!
-      email: String
+    createPersonUniqueByEmail(
+      email: String!
```
lionsdwarf: To do: apply the input type pattern to this mutation.

lionsdwarf: It's worth mentioning the duplication of input fields for this mutation and the `upsertPersonClient` mutation. I was thinking about using a schema fragment for DRY reasons (if that's even possible to use for an input type, idk, tbd 🤷‍♂), but the fragment approach is actually what made me think in more depth about the Person/Client child/sibling relationship and how that should be unwound. Which goes to say that it may be wise to be rigorous with fragments because they force you to consider where there may be some mis-modeled domain overlap. Just food for thought!

apps/api/models/client.js
craigspaeth: 👍 Nice! Mind adding this to the [README GraphQL section](https://github.com/findOctave/cortex#graphql-design) linking to those blogs :)

apps/api/models/client.js
craigspaeth: 👍I like it! While we're on the topic of the README I think we could probably break out a section about data modeling linking to our domain doc explaining our approach is to find the best midpoint between the downstream data store and our domain (e.g. we choose "person", "customer", and "user" instead of flattening to "client" because those are relevant domain names that are pragmatic to how the data is stored).

apps/api/models/person/gql.js
craigspaeth: 👍 nice, well done.

apps/onboarding/lib/view-model/consult.js
```diff
@@ -135,12 +129,26 @@ export const consultTracking = state => ({
   track: state.isCCVariation ? 'Consult Submitted 2' : 'Consult Submitted'
 })
 
+
 export const submitConsultFlow = async state => {
+  let res
+  try {
+    res = await gql.request(consultPersonMutation(state))
+  } catch (err) {
+    if (err.response.errors[0].message === 'Account already exists') {
```
craigspaeth: 👍 well done error handling to err state update here!

lib/auth/index.js
```diff
@@ -45,12 +45,13 @@ export const appendBillingLink = (e, document = window.document) => {
...
 }
 
-export const getLock = (redirectTo = '/profile') => {
+export const getLock = (redirectTo = '/profile', email) => {
   document.addEventListener('DOMNodeInserted', appendBillingLink, false)
   return new Auth0LockPasswordless(
     process.env.AUTH0_CLIENT_ID,
     process.env.AUTH0_DOMAIN,
     {
+      prefill: { email },
```
craigspaeth: Noice!

craigspaeth:
👏 Looking good! I'm fan, of all the _Scouts Ruling_ going on here! No real concerns, feels like it's moving us forward in the right direction. Well done.

lionsdwarf:
Reopening this. Changes since last WIP include: 
1) `input` type for `createPersonUniqueByEmail` mutation
2) `Client.upsertPersonClient` moved to `Person.upsertClient`
3) tests

lionsdwarf:
Closing due to new 🐛 discovered. :(   womp womp

lionsdwarf:
Bug fixed. Reopening this PR.

lionsdwarf:
Still need to update README

craigspaeth:
I left two comments I think we should address but doesn't have to block this. Well done 👏 this is great work for codebase health!

- https://github.com/findOctave/cortex/pull/976#discussion_r307902740
- https://github.com/findOctave/cortex/pull/976#discussion_r307907950

PR #977, opened 2019-07-21
Hide appointments zero state and add seed script

craigspaeth:
Just a couple of small changes I noticed and had some time to cleanup:

- Hide the Appointments section in zero state if there's no ability to schedule new appts. (edge case where a patient hasn't had a therapy/coaching/group appt. yet... the "No appointments" placeholder felt awkward and distracting without the "Schedule new appt." CTA... 🤔 come to think of it maybe a different CTA would be better like "schedule a consult to get set up with a coach or therapist"... 🤷‍♂good enough for now maybe)
- Add a database seed person table from staging to local npm script
- Run `npm test` to normalize the prettifying

cap


craigspaeth:
Tanjie and I reviewed IRL, going to merge for the seed script.

PR #980, opened 2019-07-23
End to end testing

TMcMeans:
I have some worries about whether or not line 63 is actually running, but it seems to be a perfectly working test otherwise?


test/end-to-end.test.js
```diff
@@ -49,3 +50,22 @@ describe('homepage', async () => {
...
+describe('billing', () => {
+  it.only('insert billing information successfully', async (done) => {
+    await startServer()
+    await nightmare
+      .goto(`${APP_URL}/billing?email=it@findoctave.com`)
+      .type('[class*="index__wrapper-num"] .__PrivateStripeElement-input', '4242424242424242')
+      .type('[class*="index__wrapper-expiry"] .__PrivateStripeElement-input', '10 / 30')
+      .type('[class*="index__wrapper-expiry-cvc"] label[for="cvc"] + .StripeElement .__PrivateStripeElement-input', '123')
+      .click('.button__root___2QieG')
+      .visible('.schedule-nav__root___1kvOm li:nth-of-type(2)[class*=schedule-nav__active]')
```
craigspaeth: Ah, yes. So it appears that this will [return a boolean](https://github.com/segmentio/nightmare#visibleselector) which is helpful in an `expect` block. We might want to chain wait in front of it and store the result in a variable to asserted later like

```javascript
const confirmStepIsVisible = nightmare
  .wait(...)
  .visible(..)
expect(confirmStepIsVisible).toEqual(true)
```

test/end-to-end.test.js
```diff
@@ -49,3 +50,22 @@ describe('homepage', async () => {
     stopServer()
   }, 60000)
 })
+
+describe('billing', () => {
+  it.only('insert billing information successfully', async (done) => {
```
craigspaeth: Don't forget to remove the `.only` part here otherwise it won't run the other e2e tests in production.

craigspaeth:
👏 Well done! Mostly LGTM, just those two small comments and then it's ready to merge.

TMcMeans:
Screen Shot 2019-07-24 at 12 48 38 PM

Committing current changes but we are still having a huge issue with the nightmare tests running at the same time and causing async issues. Screenshot above shows the error.

TMcMeans:
Multi-test async errors have been fixed! All tests are currently passing!

craigspaeth:
👏 Well done!!

PR #983, opened 2019-07-25
Add tracking to T&C agreements with Google analytics

TMcMeans:
Added a new reducer for tracking T&C agreement clicks. checkConsent is now called in the body of checkAndTrackConsent, which sends T&C agreement click counts to Google analytics


craigspaeth:
👏 well done!

PR #986, opened 2019-07-25
Add enforcing logic and test to T&C Agreement page for transfer clients

TMcMeans:
Added a new property to initial state called isTransferClient that is switched true when a new initTransferClient is created. Added control flow for checking if isTransferClient is true in order to keep inputs/checkboxes disabled until all form fields are validated and both T&C Agreements are clicked! 

First ticket worked on && tested independently hooray! 👩🏾‍💻🌟👩🏾‍💻


craigspaeth:
💯 👏 💯 👏 💯 👏

PR #989, opened 2019-07-29
[ Spike -- Do not merge ] Availity 

lionsdwarf:
These tiny `lib`s request an access token from Availity, then retrieve the `Configuration` for the Anthem BlueCross California "payer" record. It currently logs to the terminal the fields ~required~ for a successful form submission to Availity's [Care Cost Estimator API](https://developer.availity.com/partner/documentation#care-cost-estimator-professional) (creds in Lastpass) for Anthem BlueCross California.

availity_config_req


craigspaeth:
This is great! And a lot simpler than Kareo 😅 I'm going to close for now, but this'll be great in case we need to dig it back up shall we choose to go with Availity.

PR #992, opened 2019-07-30
Handle null for 404s in charge job

craigspaeth:
There was one other place we captured 404s and rethrew them. Given my last PR that returns nulls instead of 404s this fixes that code up.


apps/api/jobs/charge-sessions/index.js
```diff
@@ -29,11 +29,13 @@ export default () =>
         rollUpInvoices(recentAppts).map(sendInvoice)
       )
     )
-    log.success('Warnings and invoices sent. Waiting an hr to charge...')
```
lionsdwarf: Not that it really matters, but any particular reason for this deletion ?

craigspaeth: I just moved it down and split into two lines so it could log how many appts. it was about to charge.

PR #1008, opened 2019-08-05
Use ReactN provider component in view model to avoid state leakage

craigspaeth:
As described in the [ReactN docs](https://github.com/CharlesStover/reactn/blob/9821b9f45b05033a6d97e15b67183ab57862c2dd/README.md#managing-multiple-states) we should be using the [Provider component](https://github.com/CharlesStover/reactn/blob/master/Provider.md) to bind the global state to the react tree (thus one state instance per-user even server-side). As they say in their docs...

>  You may want two instances of your application to use two different global states, such as if you are concurrently server-side rendering your application for two different users at the same time. In this case, sharing the default global state with both users may create incorrect views or leak private data.

This sounds familiar 🤦‍♂

This PR updates the view model API to have two functions `model.use()` and `model.init(props)`. The later returns the provider component and the former will actually throw an error if the provider isn't used to avoid this default behavior of ReactN accidentally causing data leakage. This also comes with additional tests including an e2e test that reproduces the data leakage scenario.


TMcMeans:
Looks good, I like how we are slowly parsing out pieces of redux functionality to work on the server side.

lib/react-hooks-view-model/index.js
```diff
@@ -21,11 +21,20 @@ export const componentDidMount = fn => (
...
 export const define = (initialState, defintion) => {
-  const init = _.once(props => setGlobal(initialState(props)))
-  return props => {
-    init(props)
-    const [state, setState] = useGlobal()
-    const api = _.mapValues(defintion, cb => cb(state, setState))
-    return { state, ...api }
+  let initialized = false
+  return {
+    init (props) {
```
lionsdwarf: 👍

PR #1010, opened 2019-08-05
Fix an import from content module

craigspaeth:
A small import fix where we used to be doing a destructured import that changed to the new style of composed into a single object. TDDed this with a test as well to catch in the future.


craigspaeth:
I'd like to get a deploy out including this (https://github.com/findOctave/cortex/pull/1011) before the next billing run. So going to merge, but welcome your comments!

PR #1011, opened 2019-08-05
Redirect missing credit card to billing page

craigspaeth:
Some clients were missing credit cards despite thinking they were added to their accounts. This was due to a small bug in the new `upsertClient` resolver. Regardless, accounts missing a credit card on file that try to log in will hit an error. To handle this we redirect them to the billing page and server a separate copy module so we can say "Before you log in we need a card on file".


apps/api/models/person/gql.js
```diff
@@ -151,13 +151,11 @@ export const upsertClient = async (_root, args) => {
       patientId: patient.id
     }))[0]
   }
-  if (!person.customerId || !person.userId) {
-    const [customer, user] = await Promise.all([
-      Customer.upsert(person.email, args.stripeSource),
```
craigspaeth: After moving the input data to `args.input` this line was missed, it should have been `args.input.stripeSource`. This resolver needs a test, but I've got a ticket to refactor this GQL modeling a bit more for permissions reasons so I'll leave it for now.

apps/api/test/models/person.test.js
```diff
@@ -297,8 +297,15 @@ describe('upsertClient', () => {
         }
       }
     )
-    expect(services.stripe.customers.create).toBeCalled()
-    expect(services.auth0.createUser).toBeCalled()
+    expect(services.stripe.customers.create).toBeCalledWith({
+      email: 'test@co.co',
+      source: 'tok_mastercard'
```
craigspaeth: Updated the test to assert for that `args.input.stripeSource` part.

craigspaeth:
Going to merge to get this out before the next billing run. Welcome your comments though!

apps/api/models/person/gql.js
lionsdwarf: Ah, doh!! Good catch. Thanks, Craig!

apps/api/test/models/person.test.js
lionsdwarf: 👍

PR #1017, opened 2019-08-07
Salve results notifications guard

lionsdwarf:
Bad dummy data in the spreadsheet was causing the job to choke. This adds a guard to prevent such failures.

For some reason the diff looks noisier than it should. The only changes are the try/catch around `Client.findByEmail`, and the lodash `_.compact`

sr_notif


apps/api/test/jobs/notifications/salve-results.test.js
```diff
@@ -25,3 +33,85 @@ test('rollupToEmail converts a rollup of salve results into email obj', () => {
...
+        results: [
+          {
+            email: 'it@findoctave.com'
+          }
+        ]
+      }
+    ])
+    expect(res.length).toEqual(1)
+  })
+  test('it processes results with Clients even if another result lacks a client', async () => {
```
craigspaeth: 👍 nice tests!

craigspaeth:
LGTM! Thanks for hunting this down!

PR #1020, opened 2019-08-08
Include model design patterns in README

lionsdwarf:


craigspaeth:
I think we should mention something here about the services layer, but otherwise LGTM! 👏 good stuff!

lionsdwarf:
Good call, I meant to add that and spaced it. Updating now...

PR #1022, opened 2019-08-08
Improve Google Maps on /classes page

coltonTB:
1. Adds transit overlay to google maps
2. Disabling scrolling, zooming, and default GUI for now, since there's only one location and the existing scroll-to-zoom interaction was kind of heavy-handed
3. Remove business names

## Implementation:
Migrates from `react-google-maps` library to `google-map-react` which is a much higher quality API and actually supports `TransitLayer` (which `react-google-maps` did not)

### Before:
image

### After:
Screen Shot 2019-08-08 at 5 53 51 PM

## Todo:
Investigate color theming. (IMO the existing color scheme, while not Octave-branded, is pretty standard and does the trick). Open to discussions


coltonTB:
Also wanted to track here that the API key isn't working on my dev set up, probably because I need a .env (but wanted to flag)

package.json
```diff
@@ -57,6 +57,7 @@
     "express": "^4.17.1",
     "extract-css-chunks-webpack-plugin": "^4.5.2",
     "find-up": "^4.0.0",
+    "google-map-react": "^1.1.4",
```
craigspaeth: <3 npm... if you don't like the current module, there's probably another couple dozen options to try!

craigspaeth:
@coltonTB Well done! Agreed that the default color scheme is good. Given Octave's branding is pretty colorful and at the same time a very restrictive palette (due to our hesitance to expand branding elements until a later thoughtful identity effort), I think this is the best call.

PR #1023, opened 2019-08-09
Fix vertical alignment of inputs and hero text

TMcMeans:
Wanted to make the hero text and inputs fit together vertically to match the mockup better. Also wanted to note that the video doesn't autoplay on iPhone (and possibly iPad as well), so we might wanna do some QA on that.


craigspaeth:
👍 easy peasy!

PR #1025, opened 2019-08-09
Therapist profile modal update

TMcMeans:
Used Nextjs ```Router.push(href, href, { shallow: true })``` method to implement shallow routing on therapy page so that the provider name/credentials are displayed in the URL when their profile is clicked in the modal without triggering a full page re-direct.


craigspaeth:
w00t w00t! Good stuff 👏

PR #1026, opened 2019-08-09
[ Spike - Do not merge ] Eligible integration

lionsdwarf:
Introduces a `Coverage` model on the `Person` type that's responsible for fetching insurance coverage and eligibility information specific to an insurance plan. 

See input arguments below. 

findCoverageForPerson

Was unable to run a real (e.g. non-sandbox) coverage query against Eligible. It appears our account is stuck in sandbox mode.


apps/api/models/person/gql.js
```diff
@@ -101,7 +103,7 @@ export const toGql = person => ({
 })
 
 const people = async (_root, args, ctx, ast) => {
-  authorize(args, ctx, ast)
+  // authorize(args, ctx, ast)
```
lionsdwarf: If we maintain this schema design, we will have to address the publicity of the `coverage` resolver on `Person`. Punted this issue by commenting auth.

apps/api/models/coverage/gql.js
```diff
@@ -0,0 +1,27 @@
...
+      remainingOutOfNetwork: String
+    }
+    type Coverage {
+      isActive: Boolean
+      deductible: Deductible
+    }
+  `
+}
+
+export const findForPerson = async (_root, args, ctx) => findCoverage(ctx.variableValues.where.coverage)
```
lionsdwarf: This use of `ctx` is hacky and should be fixed. For reasons I don't understand, the args object within this field-level resolver was different from the args received in the `Person` resolver.

lib/eligible-promise.js
```diff
@@ -0,0 +1,21 @@
+import Eligible from 'eligible-node'
+
+const eligible = Eligible({
+  apiKey: process.env.ELIGIBLE_API_KEY,
+  isTest: true
```
lionsdwarf: Removing this line did not seem to toggle the sandbox environment off.

apps/api/models/coverage/gql.js
craigspaeth: Huh, odd. Happy to pair on this.

apps/api/models/coverage/mapper.js
```diff
@@ -0,0 +1,4 @@
+import eligible from 'eligible-promise'
+import { toModel, toEligible } from './domain'
+
+export const findCoverage = async insuranceInfo => toModel(await eligible.getCoverage(toEligible(insuranceInfo)))
```
craigspaeth: This function composition makes my FP heart flutter 😍

apps/api/models/person/gql.js
craigspaeth: 👍I see. We _could_ do something like we did for inspecting the fields queried... but I suspect a more elegant/safe solution would be to consider how to model the gql schemas in a way that aligns with the auth concerns.

lib/eligible-promise.js
```diff
@@ -0,0 +1,21 @@
...
+})
+
+const getCoverage = insuranceInfo => new Promise((resolve, reject) => {
```
craigspaeth: Since the eligible API uses promises this can probably be simpler as an async function...

```javascript
const getCoverage = async insuranceInfo => {
  try {
    const coverage = await eligible.Coverage.all(insuranceInfo)
  } catch (err) {
    console.error(err)
  }
})
```

That said, it seems like the eligible SDK is pretty solid and it may make sense to just simply wrap it so it doesn't complain in tests like we did for [knex](https://github.com/findOctave/cortex/blob/master/lib/db/index.js#L9)

craigspaeth:
👏 So much nicer! Now 🤞this works for the Ops team. I'm going to put a sync on Monday for you to demo me some of the software and figure out next steps about getting buy-in from Ops/getting a non-sandbox environment.

lib/eligible-promise.js
lionsdwarf: Oh duh, not sure what I was thinking regarding the promises. Sorry about that! 

Agreed.

lionsdwarf:
I like it 👍 !

lionsdwarf:
🤞🤞🤞🤞🤞

PR #1028, opened 2019-08-10
More logs in charge job

craigspaeth:
An appointment charge was mysteriously overlooked that's older than 2 weeks. I'd like to add more logs through the job lifecycle to make sure we have visibility at each step including the historical charge part.


apps/api/jobs/charge-sessions/index.js
```diff
@@ -16,7 +16,7 @@ import moment from 'moment'
 
 const oneHour = 1000 * 60 * 60
 const twoWeeksAgo = moment().subtract(2, 'weeks')
-const threeMonthsAgo = moment().subtract(3, 'months')
+const onYearAgo = moment().subtract(1, 'year')
```
craigspaeth: I noticed an appointment even older than 3 months was overlooked. It seems possible that either clinicians are forgetting to mark appts. months old and/or our historical job is timing out and leaving things behind.

craigspaeth:
Going to merge b/c changes are small and I need to get a test run on prod. going, but welcome comments!

PR #1029, opened 2019-08-11
Use day of month to generate ics file

craigspaeth:
While investigating the reminder failures I noticed an issue with the ICS file which was pulling the wrong day of month because `moment().day()` is day of week while `moment().date()` is what we want for the ICS lib (day of month).


craigspaeth:
Going to merge b/c changes are small and I need to get a test run on prod. going, but welcome comments!

PR #1030, opened 2019-08-12
Import compressed skyline video in hero and clean up waitlist inputs …

TMcMeans:
…to look better with new footer design!


craigspaeth:
Nice! LGTM!

PR #1032, opened 2019-08-12
Embed wix page

TMcMeans:
Wix page isn't enabled for responsiveness. I tried to use CSS to make surrounding container responsive but I think this needs to be a change in Wix. 

Screen Shot 2019-08-12 at 4 58 46 PM
<-------------->
Screen Shot 2019-08-12 at 5 12 50 PM


craigspaeth:
Nice! LGTM! Wix doesn't in fact handle responsiveness, it does ["Device Detection"](https://www.air-balloon.eu/device-detection-vs-responsive-web-design/) meaning you have to actually hit it with a phone instead of shrinking your browser to see the mobile version.

PR #1033, opened 2019-08-12
Fix empty el handling in content module

craigspaeth:
Short follow up to https://github.com/findOctave/cortex/pull/1024 where I noticed when an element was left blank e.g. the h2 below...

```markdown
# Hello
##
### World
```

...it would choke because the code wasn't expecting that. This was reproed on some copy that was on staging and not production which brings me back to a ticket we need to address about syncing prod to staging continuously (prioritized for next sprint).


craigspaeth:
Gonna merge and get another deploy out so we can support empty content in prod again.

PR #1035, opened 2019-08-12
do not render billing component until stripe.js script loads

lionsdwarf:
1) Do not render stripe components unless `stripe.js` loads (may now want a placeholder for this too)
2) Bug fix: credit card UI pulls from `state` instead of `client` query to avoid stale "last 4 digits" and "expiry" after cc update


apps/portal/components/settings/index.js
```diff
@@ -47,7 +47,7 @@ const StripeForm = injectStripe(props => {
         className={cx('card-submit-btn')}
         onClick={() => {
           submitting()
-          updateCreditCard(props.client)
+          updateCreditCard(state.client)
```
craigspaeth: Since you're just re-passing state in here you can actually remove this and update the reducer instead. In fact I bet there's more related cleanup to be had from this PR https://github.com/findOctave/cortex/pull/1006/files

apps/portal/components/settings/index.js
```diff
@@ -60,10 +60,40 @@ const StripeForm = injectStripe(props => {
...
+        </div>
+      )}
+    </div>
+  )
+}
+
+const Main = ({ client }) => {
   const stripe =
     typeof window !== 'undefined' &&
+    typeof window.Stripe !== 'undefined' &&
```
craigspaeth: Ah! I see what happened. So maybe Stripe is [loaded async](https://github.com/stripe/react-stripe-elements#loading-stripejs-asynchronously) or w/e is causing it to not be available right away and we were only checking for server vs client—not `window.Stripe`.

craigspaeth:
Code LGTM! However, it does point out some loose ends on cleanup. Would you mind un-skipping the tests in [this file](https://github.com/findOctave/cortex/blob/master/apps/portal/test/lib/view-model.test.js#L102) and ideally this would come with an additional test for the null soaking `&&` around `stripe`—either at an Enzyme/component level, function level, or extracted as its own function on the component or view model level.

If any of that feels unclear/arduous, feel free to make a ticket for that instead and we can address separately.

apps/portal/components/settings/index.js
TMcMeans: Is updating the reducer a better way to go with this? Should I still merge this PR or hold off for another commit?

lionsdwarf:
@TMcMeans Yes, let's pair on this during our session today.

apps/portal/components/settings/index.js
lionsdwarf: Exactly. 

I should have mentioned that, since we were unable to repro this bug, this guard was added as a precautionary measure. 

I think the most practical guard probably lives [here](https://github.com/findOctave/cortex/pull/1035/files#diff-3430b1d251fda50d3db812c68260ea7eR105), but it's really hard to know for sure without a repro.

apps/portal/lib/view-model.js
```diff
@@ -90,6 +90,10 @@ const updateCreditCard = async (state, client) => {
 export const creditCardExpirationLabel = client =>
   `Expires ${moment.utc(client.creditCard.expiresTime).format('MM/YYYY')}`
 
+
+export const isStripeLoaded = () => typeof window !== 'undefined' &&
```
craigspaeth: 👍 this is good, now can we get a test pl0x :) you can use arguments to stub for simplicity sake `isStripeLoaded = (win = window) => win.Stripe`

lionsdwarf: We opted for mocks via `global.window...`. Hope that's ok.

lionsdwarf:
Also included some cleanup around `client` prop. This is read for review again. Thx!

apps/portal/test/lib/view-model.test.js
```diff
@@ -190,3 +187,20 @@ test('addAppts changes loading state', () => {
     loadingAppts: false
   })
 })
+
+describe('isStripeLoaded', () => {
```
craigspaeth: Sweet, thanks for the thorough tests!

PR #1036, opened 2019-08-13
Refactor upsert client to handle various auth scenarios better

craigspaeth:
This consolidates `Client.updateClient` and `Person.upsertClient` into one mutation on Client. This mutations then authorizes against 3 valid scenarios to address [this comment](https://github.com/findOctave/cortex/pull/976#discussion_r307907950)...

1. Logged out creating a new "client" (e.g. the person doesn't have a customer + user yet)
2. Logged in updating your own client
3. Logged in updating a client as an admin

Did a little _Scouts Rule_-ing and adding test coverage as well.


craigspaeth:
Completely concur with what you're saying about data modeling Joe. A lot of these thoughts crossed my mind as I embarked on this too! I am starting to think marching towards more granular models that are relationships to person makes sense as much as my DDD instinct drew me in a different direction with the whole client vs. person thing.

So like `person.user` + `person.customer` vs. `person.client` or `client`. It's cheap in GraphQL to add granular relationships like that as opposed to REST, avoids over-fetchings like we currently do to get `{ person, user, customer }` for every `toClient`, and it might give us more boundaries to naturally split auth (e.g. instead of `upsertClient` it's like a `createPersonCustomer` mutation that doesn't have to check against return fields).

Let's keep this modeling discussion going as we PR or I welcome a more formal refactor ticket, docs, meetings, or something to help come to conclusion on this topic.

Good thoughts! 🙌

lionsdwarf:
Yes, great point! Maybe a Client model is unnecessary entirely, and totally agree that the granularity of these relationships will have important/helpful implications for auth. 

The march continues!

PR #1039, opened 2019-08-13
Add purchased adobe stock video and fix video attribute to work for r…

TMcMeans:
…eact. Got the official purchased video from marketing and added as a static resource. We might want to look into adding an area on airtable for switching out marketing videos to make this more seamless and less of a burden on engineering?


craigspaeth:
> We might want to look into adding an area on airtable for switching out marketing videos to make this more seamless and less of a burden on engineering?
👍 got a [ticket for that](https://trello.com/c/UumLGOOa/845-🐬-add-video-support-to-content-modules-in-airtable)

PR #1040, opened 2019-08-13
Add meta tag in _document.js to prevent search engine from indexing staging. 

TMcMeans:
Used conditional meta data tag based on app url in _document.js


craigspaeth:
Works for me! Thanks!

PR #1047, opened 2019-08-14
Wait for stripe to load in window before customer submits card info in customer-form

TMcMeans:
Similar to https://github.com/findOctave/cortex/pull/1035 where the customer billing form needed to wait for window.Stripe to load in the browser before handling user credit card data. This fix solves bugs in the e2e tests related to handling customer billing. Hopefully, all tests will pass on Circle CI now :ok_hand:


lionsdwarf:
LGTM !! 🤞

PR #1055, opened 2019-08-16
WIP - Replace `envenc` w custom lib

lionsdwarf:
Stole the code in this [solution](https://stackoverflow.com/questions/49021171/how-to-derive-iv-and-key-to-crypto-createcipheriv-for-decryption/49021891#49021891) because it conformed to my research, specifically ~[here](https://hptechblogs.com/password-hashing-in-node-js-using-the-pbkdf2-in-crypto-library/)~ and [here](http://vancelucas.com/blog/stronger-encryption-and-decryption-in-node-js/).
 
## To do:
- [x] Tests
- [ ] Evaluate level of effort in order to open-source the lib
- [ ] Consider upgrading code to ES6 syntax (e.g. `import..from`)


lib/envencdec/index.js
```diff
@@ -0,0 +1,70 @@
...
+module.exports.envenc = () => {
+  if (!fs.existsSync('.env')) throw new Error('.env not found')
+  const env = fs.readFileSync('.env', 'utf8')
+  const envenc = encryptString(env, PASSWORD)
+  fs.writeFileSync('.env.enc', envenc)
+}
+
+module.exports.envdec = () => {
+  if (!fs.existsSync('.env.enc')) throw new Error('.env.enc not found')
+  const envenc = fs.readFileSync('.env.enc', 'utf8')
```
lionsdwarf: I'd like to review/discuss the utf8 encoding here, specifically considering [the base64 encoding](https://github.com/findOctave/cortex/pull/1055/files#diff-d5f121d4481b52d53eb7a1a662ad81a9R35)

Changing this readFile to base64 encoding b0rked.

craigspaeth: I think using base64 on read will return a string in base64 encoding—essentially re-encoding it in base64. Reading the plain utf8 text will return the base64 encoded string that can then de decoded [on this line](https://github.com/findOctave/cortex/pull/1055/files#diff-d5f121d4481b52d53eb7a1a662ad81a9R40)

lib/envencdec/index.js
```diff
@@ -0,0 +1,70 @@
...
+const fs = require('fs')
+
+const ALGORITHM_NAME = 'aes-128-gcm'
+const ALGORITHM_NONCE_SIZE = 12
+const ALGORITHM_TAG_SIZE = 16
+const ALGORITHM_KEY_SIZE = 16
+const PBKDF2_NAME = 'sha256'
+const PBKDF2_SALT_SIZE = 16
+const PBKDF2_ITERATIONS = 32767
+const PASSWORD = process.argv[1]
```
craigspaeth: I think it'd make more sense to take this as arguments to the functions, e.g.

```javascript
module.exports.envenc = (password) => {
``` 

Should be easier to test that way and is similar to how [envenc](https://github.com/lapwinglabs/envenc/blob/master/index.js#L54) does it.

craigspaeth:
Nice! Seeing how randomness is used to, in a way salt, the encrypted file I'd just double-check to make sure decryption works across different machines at different times (seeing as some referenced articles talk about passwords which are meant to be one-way encrypted/compared vs. two-way encrypted/decrypted like the env file).

Otherwise this LGTM. Feel free to merge after writing a test (which I understand might be hard, so not looking for super-comprehensive here—at-least two tests: 1. check encrypt + decrypt with same password results in original text, 2. encrypt + decrypt with differing passwords results in garbage)

FYI: I noticed you checked in the resulting .env.enc file. That's okay b/c the code checks out, but you'll want to be careful about this. If there was a bug in your code the resulting file might not have been truly encrypted. So just be extra careful here—it's basically the only place in our codebase where we have to be extra careful about checking it in.

lib/envencdec/index.js
lionsdwarf: 👍 just pushed this update.

PR #1057, opened 2019-08-19
Remove robots meta tag from /marketing _documents.js file and add to /marketing/Head based on environment variable

TMcMeans:


TMcMeans:
Hopefully cleaning this up and removing the double meta tags will help Google de-index faster.

PR #1060, opened 2019-08-20
Circle CI config to push to dedicated stack

lionsdwarf:


lionsdwarf:
I think for now we should just deploy to both "production" apps.

craigspaeth:
Sounds good to me!

PR #1065, opened 2019-08-20
Support hiding location in appt confirmation emails from ICS file by using appointmentType ids in airtable

TMcMeans:
Screen Shot 2019-08-20 at 2 32 13 PM

Our solution made the location property value conditional based on apptType category. All email confirmations are "turned on" only by adding an id to the appointmentTypes column in Airtable. 

Production on Airtable has email confirmation notification settings for consults and in-person intakes. I'm not sure which ones are supposed to be virtual. Should we add appointmentTypes ids to all the relevant notifications in production? When you say "turn on" for consults & coaching do you mean add an appointmentTypes id to each of their notifications settings on Airtable? Clients won't receive notifications without it.


lionsdwarf:
Code looks good! Let's also please include a unit test for the `isVirtual` function. And on second thought...maybe that function should be called `isVirtualLocation`. What do you think ?

craigspaeth:
👏 well done—easy peasy!

TMcMeans:
I like that idea! I'll rename the function and add a unit test for it.

TMcMeans:
Unit tests are up! Also, organized the notifications.test.js file to group unit tests to make the file more readable. Let me know what ya'll think!

craigspaeth:
👏 nice updates!

PR #1072, opened 2019-08-26
Enhance notifications job

TMcMeans:
Coaching and consult phone calls are now set to a "virtual" location in email reminders based on business location (in Cliniko).


craigspaeth:
Finally approved on this!

PR #1078, opened 2019-08-29
Sync airtable Prod base > QA base

lionsdwarf:
Wholesale airtable migration of Prod data into QA tables...

However, there are related-record issues stemming from non-unique primary keys, as we are relying on airtable typecasting to manage the associations.  You can see that duplicate key values break the associations. 
Screen Shot 2019-08-29 at 1 19 26 PM

Either we need to enforce unique primary keys (if possible), or manage the remapping by relying on the ids returned from the Create calls--a non-trivial solution atop some already-difficult-to-manage code. 

Can pick back up another day and continue this slog, but I need a break from this for now.


craigspaeth:
👏 well done! This code aint the prettiest, but it's totally followable and well crafted for the nature of this complex beast. I'm +1 to enforcing unique primary keys in AT if we can do that. Definitely feel free to close this PR and pick it back up another time. This project is important but not time-sensitive, and we do have other more time-sensitive things.

From what I'm seeing I'd be happy to merge when you're ready. Ideally, we'd write at-least one integration tests here that runs the job and uses the AT API to query that a single row from prod got synced to QA—but I'm not even going to let that block this.

Exciting stuff!

lionsdwarf:
Let's review the duplicate primary key records in airtable to understand the broader implications of enforcing uniqueness. Take `notifications` table for instance: looks like we're currently relying on dupe primary key records, but I've not dug in on it...

lionsdwarf:
As discussed IRL, this PR is updated to join by prod id rather than primary key, which thankfully eliminates a bunch of nasty code. Also includes integration test.

PR #1079, opened 2019-08-29
Locations Index and Flatiron Location Pages

mkandler:
- Adds locations to the nav (requires a GQL request)
- Builds out sections for the /locations page
- Builds out sections for the /flatiron-distract location page

Wasn't sure how we want to handle URLs for the specific locations, for now just using `/name-of-location` though I think `/locations/<id>` would be better in the long run.

Also, need to move some of the content to production


craigspaeth:
👏 Looks great! Looks like you just need to revert your `package-lock.json`and/or pull the latest to resolve the merge conflict.

mkandler:
@craigspaeth - thanks for the review! Just pulled the existing package-lock.json
Good to merge? And then is there a method to get it on staging? Forgot the setup...

craigspaeth:
Merging auto-deploys to staging. So should be up soon!

PR #1080, opened 2019-08-29
Fix class booking billing bug

TMcMeans:
submitBilling(state) was using a stale copy of state where the person's id wasn't being set. Craig and I added a gql request for the person's data in submitClassFlow, updating state with fresh data from the server so person.id exists when they submit billing info for a class.


craigspaeth:
Easy peasy!

PR #1081, opened 2019-08-30
Insurance

mkandler:
Builds out the Insurance page with new components (reusing some of course)
- Also updates SlideController to have `autoPlay={false}` option


craigspaeth:
👏 nice! Thanks!

PR #1085, opened 2019-09-04
Cbrown/intake flow v2

coltonTB:
First version of static markup and style for consult flow v2. Will need some polish and TLC but hoping we can start breaking into sub-tasks from here.

image
image
image
image
image
image


craigspaeth:
Nooice! A lot to digest but a quick glance over it looks legit and consistent with a lot of the code patterns in place 👏 Well done!

PR #1086, opened 2019-09-06
Dynamic "Select Location" content from Cliniko and Airtable

lionsdwarf:
Pulls locations from Cliniko and maps the respective images from "locations" airtable.

Includes some boy-scouting cleanup around Location mapper functions, which were previously not using our "toModel" functionality, which leaked "toModel" calls into separate models (see diff for `models/appointment/gql.js`) as well as some snake_cased properties into `jobs/notifications`. (<-- There is similar cleanup to be had in the practitioners mapper functions as well, but that's a pretty far reaching cleanup effort. )

These changes could effect our `jobs/notifications` with regard to timezone, so if someone wants to take a second stab at QA-ing our client-facing notifications, I'd be much obliged.

More comments inline.


apps/api/models/location/mapper.js
```diff
@@ -1,12 +1,27 @@
...
+      if (atSibling && atSibling.images) {
+        loc.images = ContentModule.images(atSibling)
+      }
+      return toModel(loc)
+    })
+  )
+}
 
 export const findByPractitioner = async practitionerId => {
   let locations = await findAll()
```
lionsdwarf: This function should also leverage "toModel" but I did not dive into that refactor hole....so there is a little inconsistency here atm. 😢

apps/api/models/content-module/domain.js
```diff
@@ -17,7 +17,7 @@ export const airtableImgToResizeUrl = (image, width, height) => ({
   filename: image.filename
 })
 
-const images = record => ({ width, height }) => {
+export const images = record => ({ width, height }) => {
```
lionsdwarf: At some point, we may decide that Images should be it's own model.

apps/api/models/location/domain.js
```diff
@@ -1,14 +1,17 @@
...
-export const toGql = location => ({
+export const toModel = location => ({
   id: location.id,
   address1: location.address_1,
   address2: location.address_2,
   city: location.city,
   state: location.state,
   zipCode: location.post_code,
   country: location.country,
+  name: location.display_name,
```
lionsdwarf: Using the `display_name` property here vs the `business_name`, since we may decide to enforce some consistency with `business_name`. 🤷‍♀️

apps/onboarding-v2/components/location-picker/index.js
```diff
@@ -1,27 +1,35 @@
...
       Select
     </button>
   </div>
 )
 
 export default props => {
+  const { locations } = props
+  const bryantPark = _.find(locations, { address1: '286 Madison Avenue' })
+  const flatiron = _.find(locations, { address1: '45 E 20th Street' })
+  const marketStreet = _.find(locations, { address1: '635 Market Street' })
```
lionsdwarf: Referencing each location via street address here. We could also use the "business_name" but I figured this is fine...

TMcMeans:
Nice! Looks all good to me!

PR #1090, opened 2019-09-06
Onboarding availabilities

lionsdwarf:
With the new cliniko practitioner provisioned for SF, this no parses availabilities based on the selected location timezone. 

See comment inline pls.


lib/availability-picker/util.js
```diff
@@ -83,7 +83,6 @@ export const consultAvailabilities = appointmentType =>
       }))
     )
     .flatten()
-    .uniqBy(a => a.startTime)
```
lionsdwarf: I believe this was a remnant from days past when there were multiple intake coordinators at a single location and we wanted to de-dupe availabilities in the picker. Do we still need to support multiple intake coordinators per location? 

If so, we will need to re-implement this with uniqueness on `startTime && practitioner.id`.

PR #1095, opened 2019-09-09
Billing form renders for SF users, Small stripe form

lionsdwarf:
https://trello.com/c/g8tkjmuT/936-%F0%9F%90%AC%F0%9F%8C%89%F0%9F%91%8Bbilling-page-stripe-form


apps/onboarding-v2/constants/index.js
```diff
@@ -0,0 +1,5 @@
+export const locations = {
+  MARKET_STREET: '635 Market Street',
```
craigspaeth: Can we hard code business IDs instead of addresses? It seems unlikely, but if a name changes this will break.

apps/onboarding-v2/components/schedule-consult.js
```diff
@@ -42,7 +42,9 @@ export default props => {
       nextHandler={nextStep}
     />,
     <InfoForm terms={props.content.terms} nextHandler={nextStep} />,
-    <Billing content={props.content} nextHandler={nextStep} />,
+    goldenGateBridge(state) && (
```
craigspaeth: As much as this function name brings me joy, I think we should be more explicit here and name it something like `isSFLocationSelected`.`

craigspaeth:
Two small naming/config things, but overall LGTM! 👏

apps/onboarding-v2/constants/index.js
lionsdwarf: 👍

lionsdwarf:
Updates pushed!

PR #1098, opened 2019-09-10
Homepage banner

mkandler:
Adds the homepage banner and adjusts layout components
(Note: change the h1 to "hide" to hide the banner)


lib/components/nav/index.js
```diff
@@ -23,15 +23,20 @@ class Header extends Component {
...
+      bannerHeight: 0
     }
     this.scrollActivated = false
   }
 
   componentDidMount () {
-    this.setState({
-      loggedIn: Boolean(getAuth0Token())
-    })
+    if (document.getElementById('banner')) {
```
lionsdwarf: Not a big deal but I feel like this conditional value could be passed in as a prop

PR #1100, opened 2019-09-10
Complete Consult validations for credit card form and TOS

lionsdwarf:
findCoverageForPerson

Note: there is currently no CSS for button "disabled" state


craigspaeth:
Nice use of function composition and `_.get`!f

PR #1102, opened 2019-09-10
Fix notifications choking on missing practitioners

craigspaeth:
In-house confirmations were failing today because a practitioner was missing from the data. We accommodated this with a [guard conditional](https://github.com/findOctave/cortex/compare/master...craigspaeth:reminder-config?expand=1#diff-867a61e871f4b934e24b07c626301abeR49) that returned null—but after a recent refactor that `if` was located later in the code when `practitioner` was already being used therefore throwing an error that interrupted the job. This moves that if back up and adds a test.


lionsdwarf:
LGTM. Just need to fix the broken test, then ready to merge. In the meantime, I'm going to close along with the other WIP PRs.

PR #1103, opened 2019-09-11
hide new sf features behind env var

lionsdwarf:


apps/marketing/components/Providers/index.js
```diff
@@ -113,7 +113,7 @@ class Providers extends Component {
           <Reveal className={cx('header')}>
             {h1 && <h2>{h1[0]}</h2>}
             <div className={cx('filter')}>
-              {groups && (
+              {process.env.IS_SF_STAGING && groups && (
```
mkandler: I think it would be easier to include this in `therapy.js` like you did for SortableLocations. Disabling groups entirely in that way would be much cleaner and less likely to break

lionsdwarf: Thanks for highlighting and clarifying IRL. Will fix and reopen PR.

PR #1109, opened 2019-09-11
Add min-height to mobile menu so it doesn't hide from translateY 

craigspaeth:
Looks like the new `translateY` trick to our banner was hiding the mobile menu. This [SO answer](https://stackoverflow.com/questions/42398153/using-translatey-to-move-hidden-element-without-blank-space) somewhat explains the reason this happens and why we use this min-height trick to make it visible again.


craigspaeth:
Thanks @UroojNawaz this was all you! You discovered the bug in QA and pointed me in the right direction! Team work 💪 !

craigspaeth:
FWIW @mkandler I wonder if using `position: sticky` on the header part and leaving the banner `position: relative` would be a simpler approach to this than using JS to change `translateY` on scoll. Food for thought!

mkandler:
Thanks for fixing that @UroojNawaz ! Sorry I totally missed the bug in testing...

@craigspaeth definitely open to a css-only solution if that's possible. I'll take a crack at trying to make that work, sounds a lot cleaner

PR #1122, opened 2019-09-13
Support remote flow

craigspaeth:
This supports the remote location selection by using `moment.tz.guess()` to pick the timezone which to adjust the scheduler and renders "Over the phone" in the button's label. I also did more _Scouts Rule_-ing on this to remove unnecessary props passing.

cap


apps/onboarding-v2/components/schedule-consult.js
```diff
@@ -1,45 +1,30 @@
...
+    <Layout>
+      {
+        [
+          <ConsultIntro />,
+          <LocationPicker />,
+          <AvailabilityPicker />,
+          <InfoForm />,
+          <TOS />,
+          <Confirm />
+        ][state.step]
```
lionsdwarf: Nice!

PR #1125, opened 2019-09-13
Preserve Person form inputs 

lionsdwarf:
The create Person response only returns the person id, and the email is needed for the create Appointment request, so we needed to preserve the form inputs.


lionsdwarf:
Going to merge so that I can wrap up this ticket: https://trello.com/c/eGGXE0XE/940-🐬🌉👋-add-to-calendar-opens-dropdown-like-the-current-flow

craigspaeth:
👍 can you follow up with a test for this @lionsdwarf later

PR #1127, opened 2019-09-13
Fixes broken providers dropdown

mkandler:
Forgot to change this function when simplifying the identifiers to use `name` instead of `p[0]`


lionsdwarf:
Thanks Matt!

PR #1132, opened 2019-09-14
Handful of intake flow logic updates

craigspaeth:
This ties up the loose ends of Tanjie's work with a little bit of cleanup to move things to state and view model functions and wrap up tests. It also includes two other small updates on top of that:

- Filter day picker to only available days
- Confirm paragraph copy is dynamic
- Fork to in-person appt. type for SF
- Pre-pick first available day in scheduler
- Fix confirm page to look at appointment instead of availability


lionsdwarf:
Looks good minus the stray console log. Feel free to merge after removing. Thank you!

PR #1134, opened 2019-09-16
Update SF Waitlist page metaData by passing in contentModule

TMcMeans:
SF Waitlist now queries for the "metaDataSF" contentModule to replace metaData static text.


craigspaeth:
Noice! Looks like we were well prepared for this already.

PR #1137, opened 2019-09-16
Render "Unavailabilites" on Desktop

lionsdwarf:
availunavail


apps/onboarding-v2/lib/view-model/util.js
```diff
@@ -50,3 +52,24 @@ export const returnToScheduler = state => ({
...
+  const times = []
+  new Array(dailyIntakeHours).fill().map((acc, index) => {
+    const hour = index + startTime
+    times.push(moment({ hour }))
+    times.push(moment({ hour, minute: 30 }))
+  })
+  return times
+}
+
+export const isSameTime = (time, avail, TIME_COMPARE_FORMAT = 'h:mm A') => moment(avail.startTime).format(TIME_COMPARE_FORMAT) === time.format(TIME_COMPARE_FORMAT)
```
lionsdwarf: Had to hand-roll this date-agnostic time comparison, as I could not find an equivalent in momentjs.

apps/onboarding-v2/lib/view-model/util.js
```diff
@@ -50,3 +52,24 @@ export const returnToScheduler = state => ({
 
 export const findLocationById = (state, locationId) =>
   _.find(state.locations, { id: locationId })
+
+export const nineToFiveTimes = (startTime = 9, dailyIntakeHours = 8) => {
```
lionsdwarf: 9-5 is hard-coded here. We will have to re-address this should the intake coordinator hours be adjusted outside of this range.

apps/onboarding-v2/lib/view-model/util.js
craigspaeth: 👍Nice! Looks good to me (minus the small note about running prettier so this doesn't trail over 80 characters 🙃)

apps/onboarding-v2/lib/view-model/util.js
craigspaeth: 👍Thanks for noting! This is totally good for now.

craigspaeth:
👏 well done! Nice test coverage and use of view model functions. Feel free to merge when 💚

PR #1142, opened 2019-09-16
Appointment scheduler copy split

coltonTB:
Made this part of the flow dynamic depending on in-person vs over-the-phone consultation. Also modified the 'change your reservation' button and made it functional

#### in-person
image

#### over-the-phone
image


apps/onboarding-v2/components/availability-picker/index.js
```diff
@@ -5,6 +5,7 @@ import model from '../../lib/view-model'
 import { availsWithUnavails } from '../../lib/view-model/util'
 
 const cx = classNames.bind(styles)
+const REMOTE_CONSULT_LOCATION_ID = '81641'
```
lionsdwarf: We have been using environment variables for Cliniko Ids like this one. In fact, this one is already in our `.env` as `CLINIKO_VIRTUAL_LOCATION_ID`.

apps/onboarding-v2/components/availability-picker/index.js
```diff
@@ -46,22 +47,44 @@ export default () => {
...
     onSelectedDay,
+    returnToLocationPicker,
     selectedAvailabilityLabel,
     selectedAvailabilitySubLabel,
     selectedDayAvailabilities,
     selectedLocation,
     state
   } = model.use()
+
+  let consultTypeCopy, locationCopy, changeLocationCopy
```
lionsdwarf: This all looks groovy to me, so just nitpicking here:

In the most recent PRs there has been a push (no pun intended) to locate logic like this within the `view-model` and leave only view stuff within components. Seems like `view-model/copy.js` would be the best fit for this particular code. 

Totally not blocking, just wanted to highlight some of the simple code-organization stuffs we've go going on atm. 👍

coltonTB: I like this, thanks!

lionsdwarf: 🤙

PR #1143, opened 2019-09-17
Filter classes by location using the Location API

TMcMeans:


craigspaeth:
Tests are fixed in master, so mergin' thanks!

PR #1144, opened 2019-09-17
Query for locationGroups for sf page header

TMcMeans:


craigspaeth:
Simple! Same thing, remember to pull from upstream then feel free to merge when 💚

PR #1147, opened 2019-09-17
Add error handling to referral page

TMcMeans:
Added error handling to deal with "account already exists" bug on the referral page. After pairing with Joe, we realized that the "unprocessable entity" bug was being reproduced in staging due to a config issue. Staging was missing (or had an incorrect value for) the `REFERRAL_NOTE_TEMPLATE_ID` env variable.


craigspaeth:
Easy peasy then, thanks for investigating!

PR #1149, opened 2019-09-17
Onboarding desktop dummies

lionsdwarf:
Fixes the timezone issues we ran into yesterday regarding the Availability Picker and the tests.


craigspaeth:
LGTM! I like the new naming and hard-coded list—more explicit and easy to follow.

PR #1153, opened 2019-09-18
Guess zipCode for users who select remote consults

TMcMeans:
While trying to QA the info form inputs I rant into a bug while trying to click on booking a remote consultation. It looks like the user zipcode was an empty string, which caused the availability picker to break. Paired with Joe on this, and we added an option to guess the user's zipcode instead. 
Yay 🥇


apps/onboarding-v2/lib/view-model/util.js
```diff
@@ -75,7 +76,7 @@ const nineToFourThirty = [
 export const isSameTime = (dummyTime, availTime, zipCode) =>
   dummyTime ===
   momentTz(availTime)
-    .tz(zipToTz.lookup(zipCode))
+    .tz(zipCode ? zipToTz.lookup(zipCode) : moment.tz.guess())
```
craigspaeth: So we have a [view model function for this](https://github.com/findOctave/cortex/blob/cf870d221977555c9f6b6a3e5c8ee20c0303b1c0/apps/onboarding-v2/lib/view-model/consult.js#L224), maybe we can pass in `timeZone` instead of `zipCode` here to use that?

craigspaeth: @lionsdwarf I noticed that `availsWithBookings` uses [two view model functions anyways](https://github.com/findOctave/cortex/blob/5d84743398979fbce8e9a26cbfabfdf6f10db2e0/apps/onboarding-v2/components/availability-picker/index.js#L105). We could probably simplify this by making it a selector.

When using reducers and selectors you can think of `state` as a sort of combination of not only the plain data model of your view's state but also a kind of [context](https://reactjs.org/docs/context.html) that can make dependency injection/passing easy.

By this set of functions not being a state-in-state-out selector we actually ended up circumventing the logic in `schedulerTimeZone` that determines what timezone to use (based on which location you picked, defaulting to guessing).

craigspaeth:
I'm okay with merging this but it'd be DRYer to use `schedulerTimeZone`. Happy to pair on it.

craigspaeth:
Tanjie & I are going to start a new branch for this.

PR #1159, opened 2019-09-18
Spike eligible

lionsdwarf:
GQL Coverage type and query accepting inputs for patient name, dob, and member id. Octave provider NPI info stored as env var json. Anthem BCBS Id is hard-coded.


.gitignore
```diff
@@ -16,6 +16,7 @@ lib-cov
 
 # Coverage directory used by tools like istanbul
 coverage
+!apps/api/models/coverage
```
craigspaeth: Was this intentional?

lib/eligible.js
```diff
@@ -0,0 +1,6 @@
+import Eligible from 'eligible-node'
+
+export default Eligible({
```
craigspaeth: It might be better to do `export default () => Eligible` so we can avoid the whole `jest.mock('at` stuff we run into with these wrapper libs.

lionsdwarf: Oh right, I'll fix

.gitignore
lionsdwarf: Yes. Git was ignoring bc of `coverage` model name collision.

craigspaeth: ooo, good point—didn't notice the `!`

craigspaeth:
Nice! LGTM! Feel free to merge when you're ready.

PR #1164, opened 2019-09-19
Marketing footer dynamic

lionsdwarf:
This pulls the `<h1/>` and `<a/>` tags from airtable `contentModules`, and uses 'childModules' for ordering the columns of links. 

@craigspaeth I decided to reverse course on DRY-ing up the Footer gql queries once I stumbled into dependencies within the Nav component. I have a working branch with some headway, should we decide to prioritize it.


apps/marketing/pages/about.js
```diff
@@ -82,6 +84,8 @@ class About extends Component {
         }
       }`
     )
+    const { footer } = data
+    footer.childModules.reverse()
```
craigspaeth: Why not just reverse these [in the footer component](https://github.com/findOctave/cortex/pull/1164/files#diff-2e04b17b263f4779a39970dbf7b1600cR27) instead of copy + pasting this everywhere upstream?

craigspaeth:
LGTM but I think we could atleast DRY up the `.reverse()` part. Feel free to merge after that.

apps/marketing/pages/about.js
lionsdwarf: This is embarassing 🥵. Just pushed a fix.

PR #1168, opened 2019-09-20
Add html and css styling for confirmation side rail

TMcMeans:
Added static text and styling for the side rail on the confirmation page. Responsiveness and styling is still a WIP, and we will probably need to make the copy editable.


craigspaeth:
LGTM! Thanks! I have some other polish I want to do so happy to pick up responsiveness.

PR #1174, opened 2019-09-23
Inject location email reminders

TMcMeans:
This PR pulls location data and injects address into email confirmation vars. I've also written two tests to confirm the address string looks as it should in confirmation emails.


apps/api/services/notifications.js
```diff
@@ -95,7 +95,15 @@ export const bookingRollupToEmailVars = ({
     .format('MMMM Do YYYY, h:mm a z'),
   appointmentTypeName: apptType.name,
   patientPreferredFirstName: patient.first_name,
-  practitionerFullName: practitioner.first_name + ' ' + practitioner.last_name
+  practitionerFullName: practitioner.first_name + ' ' + practitioner.last_name,
+  locationAddress: [
+    location.address2
+      ? `${location.address1} ${location.address2}`
```
lionsdwarf: I feel like we should add a comma after address1, e.g. `286 Madison, 22nd Floor`

craigspaeth: I'm sure we'll get feedback from Clinicial/Ops about how to format this further—for now this LGTM!

PR #1175, opened 2019-09-23
Throw error if trying to update coupons

lionsdwarf:


craigspaeth:
Easy peasy! I could nitpick for a  test here, but this is so edge caes it's not really worth it. Thanks!

PR #1180, opened 2019-09-24
Randomize the practitioners picked in consults

craigspaeth:
To make sure one intake coordinator doesn't get swamped when there's overlapping availabilities.


lib/availability-picker/util.js
```diff
@@ -76,6 +76,7 @@ export const availabilityTimeRange = (date, now) => {
 
 export const consultAvailabilities = appointmentType =>
   _(appointmentType.practitioners)
+    .shuffle()
```
lionsdwarf: 👍

PR #1182, opened 2019-09-24
remove fixed nav-height

coltonTB:
There was no need to have a fixed navbar height which was causing slightly more padding than desired on mobile

Before:
image

After:
image


craigspaeth:
Easy!

PR #1183, opened 2019-09-24
airtable prod to qa test skipped

TMcMeans:
Skipping test as a bandaid solution until the continuous data syncing ticket is closed.


lionsdwarf:
Looks like this tests fails bc a new column was added in airtable. Voilá: https://trello.com/c/DXc0ALEi/1098-%F0%9F%90%AC-move-airtable-schema-into-airtable

PR #1185, opened 2019-09-25
Ignore Unsupported Appt Type Categories

UroojNawaz:


apps/api/models/appointment-type/gql.js
```diff
@@ -30,7 +28,7 @@ export const schema = {
   ): AppointmentType
   `
 }
-
+console.log(schema.types)
```
craigspaeth: Stray console.log here.

apps/api/models/appointment-type/domain.js
```diff
@@ -3,10 +3,15 @@ import Practitioner from '../practitioner'
 import Appointment from '../appointment'
 import { findRateByAppointmentType } from './mapper'
 
+export const supportedCategories = ['COACHING', 'CONSULT', 'CLASS', 'THERAPY']
+
 export const category = apptType => {
   if (!apptType.category) return null
   else if (apptType.category.match(/class/i)) return 'CLASS'
-  else return apptType.category.toUpperCase()
+  else if (supportedCategories.includes(apptType.category.toUpperCase())) {
```
craigspaeth: 👍 nice.

craigspaeth:
👏 well done @UroojNawaz !! Two things before this is mergeable:

1. Removing the stray console.log
2. We should write a test for this. It should be pretty [straightforward](https://github.com/findOctave/cortex/blob/master/apps/api/test/models/appointment-type.test.js#L13) and we can pair on it on Monday

UroojNawaz:
Okay!

lionsdwarf:
@UroojNawaz we can wrap this up next week. Will close this PR for now.

PR #1196, opened 2019-09-27
wait for analytics

lionsdwarf:


apps/onboarding-v2/lib/view-model/util.js
```diff
@@ -40,7 +40,7 @@ export const nextStep = state => ({
 export const trackStepAndScrollUp = state => {
   const eventName =
     _.camelCase(window.location.pathname) + 'Step' + (state.step + 1)
-  window.analytics.track(eventName)
+  window && window.analytics && window.analytics.track(eventName)
```
lionsdwarf: @craigspaeth Maybe this is the wrong approach because it's better to track none than to track some ??

craigspaeth: 👍 this will work, I was going to try [`analytics.ready`](https://segment.com/docs/sources/website/analytics.js/#ready) but this effectively achieves the same thing. Thanks!

PR #1199, opened 2019-09-27
Cc emails

TMcMeans:
This ticket def started as a little fish that morphed into a dolphin when Joe and I uncovered that the referral emails weren't being sent at all because the name variable in notificationFormula wasn't surrounded in quotes 🤦‍♀ 🤦‍♀


apps/api/jobs/charge-sessions/formatting.js
```diff
@@ -152,7 +152,7 @@ export const failureNotice = ({
   )
   if (pastChargeApptIds.includes(appt.id)) return null
   const recipient = person
-  recipient.cc = practitionerUser.email
+  recipient.cc = [practitionerUser.email, process.env.EMAIL_SUPPORT_ADDRESS]
```
craigspaeth: Craig bot notices you touched a file with "charge" in the name. This will require more than one code reviewer.

Craig the human says "LGTM"!

craigspaeth:
This LGTM but I'm going to let Joe review it to be diligent about reviewing things that touch billing-related code (even though this looks like a truly minimal zero-risk change to it). 👏 well done!

apps/api/jobs/charge-sessions/formatting.js
lionsdwarf: 👍

PR #1204, opened 2019-09-30
Split charges to separate banks based on location

craigspaeth:
This uses [Stripe Connect](https://stripe.com/gb/connect) with [destination charges](https://stripe.com/docs/connect/destination-charges) to split which bank we payout to depending on the business of the appointment in Cliniko (if the zip code is in New York then we route payouts to our NY bank, everything else goes to our SF bank).

**Don't merge yet** I'm waiting on a passport/ID photo so we can set up the banks in Connect correctly before this can go to production.


coltonTB:
Solid!

apps/api/jobs/charge-sessions/formatting.js
```diff
@@ -77,6 +78,20 @@ export const buildCost = ({ appt, apptType, billableItem, coupons }) => {
...
+ * Picks the right Stripe connect ID based on Cliniko location
+ *
+ * @param {Object} location Cliniko business
+ * @return {String} Stripe Connect account ID
+ */
+export const stripeConnectAccountId = (
+  location,
+  { STRIPE_SF_ACCT_ID, STRIPE_NY_ACCT_ID } = process.env
+) =>
+  location && zipcodes.lookup(location.zipCode).state === 'NY'
```
craigspaeth: This needs to be `location && location.zipCode &&` to prevent empty zip codes from failing.

PR #1209, opened 2019-10-02
[Don't merge yet] Charge sessions fix groups

lionsdwarf:
This fix allows iteration over all group appointment attendees, not just the first group appointment's attendees.


apps/api/jobs/charge-sessions/network.js
```diff
@@ -254,23 +254,21 @@ export const getApptsWithBillables = async (from, to) => {
...
+      groupAppts.map(async groupAppt => {
+        const { attendees } = await cliniko.find({
+          url: groupAppt.attendees.links.self
+        })
+        return attendees.map(attendee => ({
           ...groupAppt,
           ...toApptAttendance(attendee.arrived),
           patient: attendee.patient
-        }
+        }))
```
coltonTB: Is there a way to lower the complexity of this logic a bit? Also not seeing the necessity of using `_.flatten`

lionsdwarf: This function is creating an Appointment record per Attendee (of which there can be multiple) on the Group Appointment. Without `_.flatten` the function will return an array of arrays.

I'm not really seeing how to simplify the logic here, but open to suggestions.

apps/api/test/jobs/charge-sessions.test.js
```diff
@@ -822,3 +830,86 @@ test('failureNotice doesnt return for old failed charges', () => {
   }
   expect(failureNotice(rollup)).toEqual(null)
 })
+
+describe('formatGroupApptsAsIndividualAppts', () => {
```
craigspaeth: Sweet! Thank you for the extensive tests.

craigspaeth:
This LGTM! Let's hold off on merging this until we pair on it together testing against prod data with charging off.

craigspaeth:
Closing for now b/c we have some other billing changes we want to roll out carefully first. Will re-open next week. Awesome work Joe 🙌

PR #1210, opened 2019-10-02
Support unknown appointment type

UroojNawaz:
I added a test and removed stray console log


apps/api/models/appointment-type/domain.js
```diff
@@ -3,10 +3,15 @@ import Practitioner from '../practitioner'
...
 
+export const supportedCategories = ['COACHING', 'CONSULT', 'CLASS', 'THERAPY']
+
 export const category = apptType => {
   if (!apptType.category) return null
   else if (apptType.category.match(/class/i)) return 'CLASS'
-  else return apptType.category.toUpperCase()
+  else if (supportedCategories.includes(apptType.category.toUpperCase())) {
+    return apptType.category.toUpperCase()
+  }
```
coltonTB: Adding curly braces and line breaks here would be great for readability!

lionsdwarf:
@UroojNawaz BOOM--nailed it!! Thanks!

@coltonTB I too prefer the linebreaks and curly braces, but this is a non-blocking style preference. 

Merginnnn'

PR #1211, opened 2019-10-02
Restyle confirmation page links

coltonTB:
Before:
image

After:
image

Also includes minor layout and alignment tweaks


lionsdwarf:
@craigspaeth In case you have trouble deciphering the modification to the envenc ;)  , @coltonTB switched the default consult to v2.

PR #1212, opened 2019-10-02
Reset charge-sessions status in psql daily at midnight

lionsdwarf:


apps/api/jobs/status-reset.js
```diff
@@ -0,0 +1,11 @@
+import { update } from '../models/status/mapper'
```
craigspaeth: When outside the model I think it'd be nicer to use the namespace `Status.update` so it's clear—nit pick for this job but in other place it could be confusing _what_ is being updated.

apps/api/jobs/status-reset.js
```diff
@@ -0,0 +1,11 @@
...
+import * as log from 'log'
+
+const resetChargeSessions = () =>
```
craigspaeth: Nit-pick: This could live in the status model as well `Status.resetChargeSession` or since it's 1 line of code it doesn't have to be its own function.

craigspaeth:
LGTM! I'll let you decide if you want to address my nitpicks and if you want to make sure a 2nd code review happens b/c while it doesn't touch billing code it does create a side-effect to it (although very minor/safe).

lionsdwarf:
@craigspaeth All good points. I'll revise...Thanks.

PR #1230, opened 2019-10-07
Change selectedDay state to day-only format string

craigspaeth:
Fixing our issue with seeing weird places where availabilities were for the wrong day.


lionsdwarf:
Feel free to merge once build is 🍏

PR #1232, opened 2019-10-07
Turn off $1 charge emails

lionsdwarf:


craigspaeth:
LGTM! Touches billing code but not the risky kind, so mergin'

PR #1235, opened 2019-10-08
Notifications dollar charge

lionsdwarf:
Let's try this and see if it will override the Customer email. Unfortunately [the docs do not specify](https://stripe.com/docs/api/charges/create#create_charge-receipt_email) a solution here.


craigspaeth:
🤷‍♂ worth a shot. Feel free to merge if you think testing in master is necessary but you should be able to tests this locally with a test CC # in the SF book consult flow or /billing page.

lionsdwarf:
As mentioned IRL, Stripe will not send notifications for test API credentials. Merginnnn

PR #1240, opened 2019-10-08
Remind appts hourly 60 hrs in advance

lionsdwarf:


craigspaeth:
👍Nice looks like we did come to the conclusion that this is meant for reminding about the cancellation window. Glad this resulted in a simple config change.

lionsdwarf:
@craigspaeth That's correct and Max was very helpful in getting Sejal on the phone to help refine the design/UX to accommodate the 48-hour cancellation window.

PR #1242, opened 2019-10-09
Feature flag for hourly notifications

lionsdwarf:


craigspaeth:
Nice! Good idea to let this bake under a feature flag first. Just be sure to make a note to clean it up once it's stable.

PR #1243, opened 2019-10-09
Update provider book button & copy text 

TMcMeans:
Added the ability for marketing to edit the cta copy text and button link with a contentModule.


apps/marketing/pages/therapy.js
```diff
@@ -166,6 +171,8 @@ class Therapy extends Component {
           {...providers}
           groups={providersGroups.childModules}
           limit={4}
+          CTACopy={providerCTA.p[0]}
+          CTAUrl={providerCTA.href[0]}
```
lionsdwarf: Let's stick to `camelCase` property names, here, please. Also, I think it makes sense to use `_.get` here so that the app doesn't b0rk if someone removes the `p` or `href` in airtable, á la `ctaCopy={_.get(providerCTA, 'p[0]')}`

TMcMeans: Gotcha, getting right on to updating that!

PR #1246, opened 2019-10-10
Add string literals to template strings

TMcMeans:
Please double check that I'm using the string literals correctly 😅the test suites are passing but I want to make sure I didn't introduce more bugs by messing with the AT calls.


apps/api/jobs/charge-sessions/network.js
```diff
@@ -118,7 +118,7 @@ export const fetchNotificationContent = async () => {
   })
   const emailFooterContent = (await at.find({
     table: 'contentModules',
-    filter: 'name = "emailFooter"'
+    filter: `name = 'emailFooter'`
```
craigspaeth: 👍 this is good. You may also use double quotes here b/c you're not interpolating a variable `"name = 'emailFooter'"`

apps/api/jobs/etl/sync-appointment-types.js
```diff
@@ -8,7 +8,7 @@ import * as log from 'log'
 
 export const toUpsert = apptType => ({
   table: 'appointmentTypes',
-  filter: `clinikoId = ${apptType.id}`,
+  filter: `clinikoId = '${apptType.id}'`,
```
craigspaeth: 👍 Looks like this was the only place that needed updating but I bet it didn't cause any problems because clinikoId is a number (although Cliniko is changing this in the future to a string, so we'll be prepared!).

craigspaeth:
Good stuff! FWIW Refactoring some of the strings to template literals was unnecessary b/c they weren't injecting variables, but I'm cool with the added consistency.

PR #1247, opened 2019-10-10
Reorganize submitConsultFlow to create person first

TMcMeans:
to create person first, then submit billing and create appt last. Paired with Craig on this and we tested it! It works 👍so that users won't schedule an appointment in Cliniko if their card is declined.


craigspaeth:
@lionsdwarf It does not update those columns but that's okay b/c when you put in a working CC and submit again it runs `upsertClient` again—creating that linkage. Turns out this was a super simple fix to just reorder two lines 🎉

PR #1248, opened 2019-10-10
Support Firefox mobile

craigspaeth:


lib/components/outdated-browser/index.js
```diff
@@ -7,7 +7,7 @@ export default () => {
     outdatedBrowserRework({
       browserSupport: {
         Chrome: 70,
-        Firefox: 60,
+        Firefox: 10, // Firefox Mobile chokes on versions 50+
```
lionsdwarf: Why 10 here ?

craigspaeth: Because 50 didn't work and since FF is on version 70 now I didn't feel like going down every version number until I figured out what the cutoff is for FF mobile. (Firefox users are pros right? If someone is using FF I doubt they have an old version 🙃)

lionsdwarf: That logic works for me! Thanks for the explanation.

PR #1252, opened 2019-10-11
Fixes for Oct 10 hard-coded dates

lionsdwarf:
Bc we are now at Oct 11, tests were failing in prod and master.


craigspaeth:
Thanks for the quick fix!

PR #1253, opened 2019-10-11
EmailCheck

TMcMeans:
https://github.com/pensierinmusica/email-check

Imported the emailCheck lib to implement email validation on the backend in the Patient model.


apps/api/test/models/patient.test.js
```diff
@@ -173,3 +173,12 @@ describe('toClinikoAddress', () => {
     expect(Patient.toClinikoAddress(input)).toEqual(output)
   })
 })
+
+describe('validateEmailExists', () => {
+  test('validateEmailExists throws error if email is not valid', async () => {
+    const result = Patient.validateEmailExists('tmoody@gm.com')
+    await expect(result).rejects.toEqual(
```
craigspaeth: Nice test! It'd be ideal to use a mock here though so we don't have to actually run `emailCheck` on every test run—resulting in faster and more predictable tests.

e.g.

````javascript
jest.mock('email-check', () => jest.fn().mockRejectedValue(new Error('refuse')))
````

File: apps/api/models/patient/domain.js [hunk 2]
```diff
@@ -98,3 +103,17 @@ export const sfWaitlistNoteToTreatmentNote = ({ patient, note }) => ({
...
   }
 })
+
+export const validateEmailExists = async email => {
+  if (email) {
+    try {
+      await emailCheck(email)
+    } catch (error) {
+      if (error.message === 'refuse') {
+        throw new Error("Email address doesn't exist!")
```
craigspaeth: Nit-pick: Since this error bubbles to the user maybe we can drop the exclamation point to be slightly less scary 🤷‍♂

craigspaeth:
LGTM aside from the one comment about using a mock in the test. Feel free to merge after that! 👏 well done!

apps/api/test/models/patient.test.js
TMcMeans: Sure! It definitely makes sense to mock out email check for testing 💯

PR #1257, opened 2019-10-14
Whitelist Stripe errors to avoid accidental failed charge emails

craigspaeth:
This simply modifies our `failureNotice` function to return null if the error isn't the card failed [type](https://github.com/stripe/stripe-node/wiki/Error-Handling) `StripeCardError`.


craigspaeth:
Thanks @lionsdwarf ! @TMcMeans @coltonTB I'm going to be diligent to get a 2nd code review as this touches billing code before merging. No pressure to be able to scrutinize every detail of how this works—just want a 2nd pair of eyes to catch silly mistakes.

TMcMeans:
LGTM as well! Did you double check your env variables?

craigspaeth:
Yep! I ran the job on staging with fresh env vars and didn't add any new one here. Thanks for the double check Tanjie. Mergin'

PR #1265, opened 2019-10-17
Fix ghost coupons

lionsdwarf:


apps/api/models/coupon/mapper.js
```diff
@@ -10,7 +10,12 @@ export const findFromCodes = async couponCodes => {
     table: 'couponDiscount'
   })
   return _.compact(
-    coupons.filter(coupon => couponCodes.includes(coupon.couponCode))
+    coupons.filter(
```
craigspaeth: Nit-pick but this logic could be split out into its own pure domain function which would make the testing easier too.

craigspaeth:
Thanks for the quick fix!

PR #1269, opened 2019-10-18
Scale down banner font for temp fix to copy overflow

TMcMeans:


craigspaeth:
Works for me, thanks!

PR #1273, opened 2019-10-18
Team mini profiles

TMcMeans:
Injected the providerCTA into midtown and flatiron pages as well. Did a scan for anywhere a Providers component is used so we shouldn't have more UI hiccups on the FE with regards to provider/coaching/team profiles.


craigspaeth:
LGTM! Thanks!

PR #1276, opened 2019-10-21
Split confirmation aside

TMcMeans:
I split the aside copy into two contentModules and query for them in initConsult(), then used isLocationSanFran() to determine which contentModule to use to render in onboarding-v2 confirmation. Ops should be able to edit confirmation aside copy based on location in airtable now. Let me know if anything needs to be refactored here!


craigspaeth:
Easy peasy! Thanks!

PR #1279, opened 2019-10-22
Use new icon and update CSS for couples therapy button

TMcMeans:


apps/onboarding-v2/components/location-picker/index.js
```diff
@@ -68,13 +68,15 @@ export default () => {
           </div>
         </div>
         <div className={cx('remote-consult-aside')}>
-          <h5>Here with your partner?</h5>
+          <h5>Here with your partner(s)?</h5>
           <div
```
lionsdwarf: Love it!

PR #1284, opened 2019-10-24
Improve legibility of MBC MVP chart

craigspaeth:
Clinical design team needs the chart to be more legible—this does that with some simple typography and design tricks like when numbers are close they sit in different corners that don't overlap.

image


apps/dass/index.js
```diff
@@ -5,11 +5,62 @@ import qs from 'querystring'
...
+        {val}
+      </text>
+    </>
+  )
+  const label = (
+    <text x={x} y={y} dy={-4} style={labelStyles} textAnchor='end'>
+      {name}
+    </text>
+  )
+  return x <= 30 ? (
```
lionsdwarf: Oooh well done! 👏

apps/dass/index.js
```diff
@@ -32,7 +83,7 @@ const buildChart = plot => {
       <Line
         type='monotone'
         dataKey='a'
-        stroke={'darkcyan'}
+        stroke='darkcyan'
```
lionsdwarf: Nit pick: the strokes for the Lines could also pull from the (`colors`)[https://github.com/findOctave/cortex/pull/1284/files#diff-b8e38778452128f1979996d6aec18857R11] object, if it was moved into the parent scope.

apps/dass/index.js
```diff
@@ -53,8 +104,20 @@ const buildChart = plot => {
...
+  )
+  $('.recharts-cartesian-grid-vertical line').attr(
+    'style',
+    'stroke: rgb(204, 204, 204)'
+  )
+  $('.recharts-cartesian-axis-line').attr('display', 'none')
+  // Shift the svg to the left a bit and return HTML string
   return $(svg)
     .attr('xmlns', 'http://www.w3.org/2000/svg')
+    .attr('viewBox', '-30 0 500 335')
```
lionsdwarf: SVG master!

apps/dass/index.js
```diff
@@ -53,8 +104,20 @@ const buildChart = plot => {
   $(lines[10])[0].attribs.stroke = 'firebrick'
```
lionsdwarf: Unfortunately these color assignments are producing an inconsistent output. Probably because these lines are ordered differently. Not sure if you know a better approach ? Should look like this:

image

lionsdwarf:
Hmmm circle ci failing on `npm install`....

PR #1288, opened 2019-10-25
Build cal content by apptType category and location

TMcMeans:


apps/onboarding-v2/lib/view-model/consult.js
```diff
@@ -200,6 +200,17 @@ export const initLocations = async state => {
   return { ...state, locations, disabledMsg: null }
 }
 
+export const initAppointmentTypes = async state => {
+  const gql = new GraphQLClient(process.env.APP_URL + '/api')
```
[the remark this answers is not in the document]
TMcMeans: Using an existing envar so no need to run this

apps/onboarding-v2/test/lib/view-model/consult.test.js
```diff
@@ -775,3 +775,43 @@ describe('onSelectedDay', () => {
     expect(result.selectedDay).toEqual('2001-09-01')
   })
 })
+
+describe.only('initAppointmentTypes', () => {
```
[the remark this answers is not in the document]
TMcMeans: Removed!

lib/ics-omorphic/index.js
```diff
@@ -0,0 +1,31 @@
...
+        apptTypeName,
+        practitionerName,
+        address1
+      ),
+      'Psychologist Therapy Session NY': Templates.buildNYTherapyCalContent(
+        apptTypeName,
+        practitionerName,
+        address1
+      )
+    }[apptTypeName + ' ' + state] ||
```
lionsdwarf: I don't think we can't rely on the conjunction of `ApptType.name` and `Location.state` here, because the Appt Type name could change. Can we use `ApptType.id` instead , or is it not a 1:1 mapping of Appt Type <> Template. We may need to join on `ApptType.id` and `Location.id`. Either way, we will want to use environmen variables for the ids.

lionsdwarf:
Left one comment but this seems to be on the right track -- bravo!

lib/ics-omorphic/index.js
lionsdwarf: Or maybe we join on `ApptType.category` and `Location.id`. You'll have to decide based on the relationship btwn templates and Appt Types.

TMcMeans: It definitely makes sense to not use strings to map templates to appt types. I think trying to implement ids would be a great way to keep it consistent in case apptType names change. Should be an easy refactor. Thanks for taking a second look! 💯

lib/ics-omorphic/index.js
```diff
@@ -0,0 +1,56 @@
+import Templates from './templates'
+
+export const buildCalContent = (
```
craigspaeth: Love this module and especially its name! 💯 for ICSomorphic!

lib/ics-omorphic/templates/consult.js
```diff
@@ -0,0 +1,73 @@
...
+export const buildMidtownConsultCalContent = (
+  appointmentTypeName,
+  practitionerFullName,
+  locationAddress
+) => ({
+  title: `Octave ${appointmentTypeName} with ${practitionerFullName}`,
+  location: `${locationAddress}`,
+  description: `
+You are scheduled for your ${appointmentTypeName} with ${practitionerFullName}. Make sure to fill out these forms at least 24h before your consultation: 
+http://www.findoctave.com/consent-treatment-NY
```
craigspaeth: 👏 well done designing this modules!

lib/ics-omorphic/test/index.test.js
```diff
@@ -0,0 +1,167 @@
...
+
+    expect(result.title).toEqual('Octave Coaching Session with Megan')
+    expect(result.location).toEqual('Virtual')
+    expect(result.description).toMatch(
+      'You are scheduled for your Coaching Session with Megan.'
+    )
+    expect(result.url).toEqual('https://www.findoctave.com/coaching')
+  })
+
+  it('buildCalContent return default copy for other appointment Types', () => {
```
craigspaeth: Nice tests! 💪

craigspaeth:
👏 Great work!!

PR #1294, opened 2019-10-30
[Do not merge] dynamic redirects

lionsdwarf:
Unfortunately using middleware for dynamic redirects guarantees that the async request to airtable always fires, regardless of the route. So this approach isn't going to cut it. 

So maybe we need to prefix redirect routes and use `app.get` instead of middleware?? 🤷🏽‍♂️


apps/seo/lib/redirector.js
```diff
@@ -4,16 +4,14 @@
...
-const app = express()
-
-const loadRedirects = async () => {
+export default async (req, res, next) => {
   const redirects = await at.find({ table: 'redirects' })
-  redirects.forEach(({ path, toUrl, statusCode }) => {
-    app.get(path, (_req, res) => res.redirect(statusCode, toUrl))
-  })
+  const redirect = _.find(redirects, { path: req.url })
+  console.log('>>>>>>><<<<<<<<', redirect)
```
lionsdwarf: This always logs

index.js
```diff
@@ -76,9 +73,13 @@ export const main = async () => {
   const nextApps = await loadNextApps()
   nextApps.forEach(nextApp => app.use(nextApp))
 
-  // API, 404, and error handler
+  // API
   app.use(reloadableApp(path.resolve(__dirname, 'apps', 'api')))
+  // SEO app with 301 redirecting, robots, etc.
+  app.use(reloadableApp(path.resolve(__dirname, 'apps', 'seo')))
+  // 404 handler
```
lionsdwarf: Order here does not matter bc...wulp...middleware.

apps/seo/lib/redirector.js
coltonTB: If we need runtime dynamic redirects, there's really no way around this. Since it looks like `caching-fetch` is being called underneath `at.find`, think we're okay

lionsdwarf:
@coltonTB This looks like it might be worth trying: https://github.com/expressjs/express/issues/1765#issuecomment-377987574

What do you think?

index.js
craigspaeth: Huh? Order should matter in middleware—that's kind the whole point of middleware is to be able to stack it in order so you can say put a 404 handler to catch everything at the "bottom".

Curious to hear what's the concern here.

craigspaeth:
@lionsdwarf You are right that at a certain point we can't avoid making the airtable fetch, but if stacking the middleware so that part happens JIT and as Colton points out, with caching it shouldn't be a big deal.

At a quick glance, this is mergeable to me. As long as the order is like this...

```javascript
app.use(nextApps)
app.use(api)
app.use(anyOtherImportantAppMiddleware)
app.get('*', redirects) // All the known routes are above to avoid this fetch
app.get('*', 404handler) // Finally a 404 unavoidably has an extra fetch but who cares
app.use(errorHandler) // Once error throws it skips the stack probably avoiding the fetch
```

lionsdwarf:
Sounds good. Thanks for the comments! I will clean this up (remove the `console.log`).

PR #1297, opened 2019-10-31
Do not drop columns Person.name, Person.email

lionsdwarf:
Our next Deploy contains some drastic changes to the notification jobs infrastructure and to foundational model code. As such, this PR is a prophylactic of sorts; if a rollback becomes necessary, the suffering will not include db schema pains.

PS I just looked through the backlog and am not sure what happened to the "Script to rollback a bad migration ticket". 🤷‍♀️


craigspaeth:
🤔 I think this can work fine because our code won't choke if the columns are still there. Good thinking to stagger the changes out! Here's that card btw, still in the backlog https://trello.com/c/48WN0RbE/573-%F0%9F%90%B3-rollback-script

PR #1298, opened 2019-10-31
Remove email check lib

lionsdwarf:
This email check code is breaking Scheduling and Billing pages, so this takes it out until we can find something more reliable.


craigspaeth:
Damn, fair enough we can remove. I'm curious to know which emails it breaks for though because we match against the `refuse` error code and swallow everything else.

craigspaeth:
@lionsdwarf Don't be shy about `npm uninstall`ing the module too btw.

PR #1314, opened 2019-11-05
Schedule Class: Use default content module for billing if cali zip code

lionsdwarf:
The `/schedule-class` flow b0rked on CA zip code because the `<CustomerForm/>` content was zip-code based and, unlike in `/schedule-consult`, we don't have a separate billing content module for Cali. This uses the default content module if scheduling a class.


craigspaeth:
Easy peasy, thanks!

PR #1320, opened 2019-11-06
[ Do not merge - code spike] Yesterdays charges resolver

lionsdwarf:
Screen Shot 2019-11-06 at 3 01 21 PM

1) Copies the core functions in `charge-sessions` into `services/billing` for use in a resolver
2) Implements a `Charge` gql type containing partial data according to Max's mockup
3) `yesterdaysCharges` query resolver that returns charge aggregates + stats such as req <> res query duration and appointments processed counts

Next steps -- Architecture decisions:

(Please note: these are end-of-day, lazy-brain ruminations)

Option 1: Run with the existing spike implementation.

Option 2: Run with the existing spike implementation + refactor `services/billing` to leverage existing model code only (it currently uses a hodgepodge of model code and direct lib network callouts).

Option 3: Implement Option 1 within a new gql API. The thinking behind this approach is that the `Charge` type should not be first class but would ideally live as something like `Appointment.charge`. If we decide to cut corners instead, a new API would provide a conceptual boundary. Also, the hodge-podgieness of the `services/billing` could also benefit from the conceptual boundary (although we would have to decide what to do about the existing `apps/api/models` usage--is it time for models to become individual `libs`?? 🤷🏽‍♂️).

Option 4 (double gold star): Implement `Appointment.charge`. This would entail some effort but would modify our schema in the correct way (instead of the corner-cutting Options 1 & 2). This could also have some nice easy wins down the road for, say, an Upcoming Charges widget on the Profile page (transparency is something we value at Octave).

Current thoughts: Options 1 or 2 are probably most reasonable in order to ship something within a reasonable time frame. 

Option 3 only seems to have conceptual benefits that hold us over until we reach greener pastures, but it may give us a nice separate of concerns for now if we can find an answer to the `apps/api/models` question. 

Final thought: If we decide that Option 4 really is the double-gold star, it may be worthwhile to timebox a day to spike it. I think/hope the modularity (🍿 to us) of our `services/billing` code will lend itself easily to generating the `Appointment.charge` object per appointment--we may only need to add an appointment id to the cliniko query. But the future here is hazy: do larger data/data-model question marks emerge? Is 'authorization' a beast? More `toModel` concerns in this approach as well. 

That's all for now 🙃


craigspaeth:
Wow! Impressively fast refactor @lionsdwarf let's timebox a full day on option 4 and if it becomes an obvious large can of worms—proceed with shipping option 1 (then down the line we'll _Scout Rule_ towards option 2 or 4). 👏 🍿 👏 🍿 👏 🍿

PR #1322, opened 2019-11-07
Delete old consult

TMcMeans:


craigspaeth:
This is a good start! I think there's more stuff to clear out in the view model, but I'll merge this as a good first step. Thanks!

index.js
```diff
@@ -45,7 +45,6 @@ const loadNextApps = async (specifiedApp = process.argv[2]) => {
 
 export const main = async () => {
   const { APP_URL, PORT, SENTRY_DSN } = process.env
-  const IS_CONSULT_V2_ACTIVE = process.env.IS_CONSULT_V2_ACTIVE === 'true'
```
craigspaeth: Maybe we can remove this from the .env.enc if it's still in there

PR #1323, opened 2019-11-07
Add link to keepingtempo.com in marketing navbar

TMcMeans:


lib/components/nav/index.js
```diff
@@ -179,6 +179,11 @@ class Header extends Component {
               </div>
             </div>
           )}
+          <Link href='https://www.keepingtempo.com' prefetch>
```
craigspaeth: `<Link>` components are a fancy way for Next.js to do client-side page transitions—we don't actually want it in this case because it won't work on external sites. Just the `<a>` is good.

craigspaeth:
LGTM minus the <Link> component!

lib/components/nav/index.js
TMcMeans: I thought to remove the Link component but was worried about styling inconsistencies

craigspaeth: There's no classnames on the link components (not sure it even shows up in the DOM), so I'd be surprised it affect styling, but if it does we may need to refactor the CSS to target the child `<a>` tag.

PR #1327, opened 2019-11-08
Prompt login in booking flows for existing accounts

craigspaeth:
This adds a login modal mid-booking-flow so we avoid the #help reports about missing zip codes from lead gen efforts.

cap


apps/api/index.js
```diff
@@ -100,7 +101,7 @@ class Logger {
       response: graphqlResponse,
       authorization: this.headers
     }
-    log.msg(`GQL: ${JSON.stringify(res)}`)
+    if (NODE_ENV === 'production') log.msg(`GQL: ${JSON.stringify(res)}`)
```
lionsdwarf: Yay!

apps/api/models/person/gql.js
```diff
@@ -116,6 +119,7 @@ const updatePerson = async (_root, args, ctx, ast) => {
   }
   person = _.pickBy(person, _.identity)
   await update(person, { id: person.id })
+  await Patient.upsert({ ...args.input, email: person.email }, 'email')
```
lionsdwarf: Do we want the `patientId` included in the Person `udpate` here ?

apps/onboarding-v2/lib/view-model/consult.js
```diff
@@ -1,15 +1,18 @@
 import _ from 'lodash'
-import { consultAvailabilities } from 'availability-picker'
 import {
-  gql,
+  consultAvailabilities,
+  createAppointment,
+  createPerson,
+  submitBilling
+} from 'availability-picker'
```
lionsdwarf: Nice

apps/api/models/person/gql.js
craigspaeth: Good point, this should be updating by patient ID not email address. Fixing!

craigspaeth:
Addressed comments and added tests.

craigspaeth:
Just noticed a not-so-edge case when a "person" exists but not a "client". Closing until I find a solution to that.

craigspaeth:
Ok, phew 😅 Not ideal but it appears that Auth0 forces a newly created account to verify their account in a redirect flow—meaning the screencap above won't work and we'll have to restart them in the beginning of the flow after logging in. A UX improvement down the line here is we can serialize the state into the url to save the spot.

To enable this I allowed for clients to be created without a `stripeSource`—meaning when a person already exists from e.g. lead-gen we...

1. Create the client without a CC to reserve the user account
2. Prompt log in
3. Restart them in the flow now authorized
4. Send authorized Person.update and Client.upsert when booking

This also further highlights to me the need for more granular modeling around people/users/patients/customers/etc. but this will get the job done for now.

lionsdwarf:
Not pretty but LGTM. Thanks! 👏

PR #1329, opened 2019-11-11
Fix log in button in nav

craigspaeth:
D'oh I changed the API to our lock lib in my last PR and introduced a breaking change. This fixes that with a test.


lib/auth/index.js
```diff
@@ -48,7 +48,7 @@ export const getLock = ({
   redirectTo = '/profile',
   email,
   instructions = passwordlessEmailInstructions
-}) => {
+} = {}) => {
```
craigspaeth: Introduces a default argument of an empty object to avoid the `redirectTo` of undefined error.

lionsdwarf: 👍

PR #1332, opened 2019-11-11
fix for toModel

lionsdwarf:
Changes in the `toModel` code introduced this bug in the billing job. Hate to say it, and maybe it's just the pain of inconsistency in the codebase, but maybe not: I'm starting to regret championing the `toModel` pattern.


craigspaeth:
Thanks for the quick fix! Low-level breaking changes are bound to break things upstream. That's just the nature of the change—not necessarily a design flaw. I'm happy to keep pushing the `toModel` stuff, if you do feel strongly we steer away though—let's make that call sooner than later b/c it'll be another painful reverse once we're further down the path.

Code LGTM. Thanks!

PR #1333, opened 2019-11-11
Charges resolver and mutation

lionsdwarf:


apps/api/models/charge/gql.js
```diff
@@ -0,0 +1,129 @@
...
+    type ChargeResult {
+      appointment: Appointment
+      error: String
+    }
+  `,
+  queries: `
+    charges(where: ChargesWhereFilter): [Charge]
+  `,
+  mutations: `
+    charge(input: [ChargeInput]): [ChargeResult]
```
craigspaeth: Should this be a list `[ChargeInput]` if so maybe the mutation should be called `charges` plural?

apps/api/models/charge/gql.js
```diff
@@ -0,0 +1,129 @@
...
+    }
+    input ChargeInput {
+      amount: Int!
+      person: ChargePersonInput!
+      appointment: ChargeAppointmentInput!
+      customer: ChargeCustomerInput!
+    }
+    type ChargeResult {
+      appointment: Appointment
+      error: String
```
craigspaeth: GraphQL already has a convention for returning errors in an `errors` response body. Let's leverage that and `throw` inside the resolver instead. (Unless I'm missing something about what `error` means in this case—in which case maybe we name it something else like `warning` or `issue` or `flag`).

apps/api/models/charge/gql.js
```diff
@@ -0,0 +1,129 @@
...
+    startTimeTo
+  )
+  return charges.map(toGql)
+}
+
+const charge = async (_root, args, ctx) => {
+  authorizeAdmin(ctx)
+  const charges = await Billing.createCharges(args.input)
+  return charges.map(resultToGql)
+}
```
craigspaeth: Yay! This file brings me joy. So nice to have a greenfield model where all of our solified patterns can create clean GQL design and model layer code.

apps/api/models/coupon/gql.js
```diff
@@ -16,7 +19,7 @@ export const schema = {
 }
 
 export const coupons = async (_root, args, ctx, ast) => {
-  Person.authorize(args, ctx, ast)
+  authorize(args, ctx, ast)
```
craigspaeth: Nit-pick: I think it'd be cool to have a convention where we keep the namespace for external model/services to make it obvious what's within the model e.g. `Auth.authorize`.

apps/api/services/authorization.js
```diff
@@ -0,0 +1,29 @@
+import _ from 'lodash'
+
```
craigspaeth: Love that you pulled this into its own service! 💯

apps/api/services/billing/formatting.js
```diff
@@ -0,0 +1,127 @@
+import _ from 'lodash'
+import { isChargableAppt } from './validation'
+import moment from 'moment'
+import momentTz from 'moment-timezone'
+import zipcodes from 'zipcodes'
+
```
craigspaeth: Well done extracting to the service layer 👏

craigspaeth:
😭 it's so beautiful. 👏 👏 👏 👏 

I only have a couple of nit-picks that you're welcome to address or not. It looks like you addressed this but I just want to doubly flag that we should be careful to make sure the old billing job code remains and can run so we can beta test this panel before we kill the job.

So excited!!!

apps/api/models/coupon/gql.js
lionsdwarf: Good call! I'll revise accordingly.

lionsdwarf:
Yes, thanks for highlighting--the job code was only copy pasta-d manually.

apps/api/models/charge/gql.js
lionsdwarf: The thinking here is that, from the bulk transaction response, we will want to map in the UI if any individual charge transactions failed. Is there a way to leverage the native error response body alongside an appointment id ? We could alternatively abandon the bulk transaction. Or as you suggest, we could rename it.

apps/api/models/charge/gql.js
lionsdwarf: 👍 It should be charges plural--thanks for highlighting!

apps/api/models/charge/gql.js
craigspaeth: 🤔 I see. I think it'd be okay to drop the bulk transaction and do single mutations (I think you could even build a GQL query of multiple mutations in the same request if you wanted to effectively get the best of both worlds).

lionsdwarf: Oh yea this is definitely the way to go. Good call!

lionsdwarf: Unfortunately this approach is not working for me: https://stackoverflow.com/questions/50889548/how-can-i-run-one-mutation-multiple-times-with-different-arguments-in-one-reques. Can give more details IRL if necessary. It might also be a bit cumbersome to generate the bulk mutation string/inputs....

In any case it's not really a pressing optimization atm.

lionsdwarf:
Nit-picks resolved so I'm merginnnn....

apps/api/models/charge/gql.js
craigspaeth: 👍 sounds good! FWIW it would be totally cool to send a bunch of roundtrips/parallel requests on the client-side of single mutations too.

The simplest/cleanest/most-consistent implementation should always trump performance optimizations that create anti-patterns or more complex code. Not saying that's the case here, but just a friendly reminder of how to think about things in the future.

craigspaeth: EDIT: I saw the outdated flag here and noticed you did indeed move it to singular and mentioned "not really a pressing optimization atm"—so 💥 you killed it! Well done on navigating tradeoffs/premature optimization 👏

PR #1337, opened 2019-11-13
Support Flatiron in booking flow

craigspaeth:
Adds the choice to book Flatiron but supporting a couple of small items pictured below. I also did some _Scouts Rule_ing and cleaned up a relevant TODO in the model layer.

image

cap


craigspaeth:
@lionsdwarf Actually, to my surprise—someone already anticipated this and added `CLINIKO_FLATIRON_LOCATION_ID` to the envenc! So should be good to merge.

PR #1340, opened 2019-11-18
[WIP] Sub-state split per page 

lionsdwarf:
Putting this here for review of per-page slicing of global app state.


lib/react-hooks-view-model/index.js
```diff
@@ -20,11 +20,13 @@ export const componentDidMount = fn => (
   }, [])
 }
 
-export const define = (initialState, defintion) => {
+export const define = (state) => {
```
craigspaeth: This'll be a breaking API change across all view models. I also personally think separating the arguments is important to make the `initialState` implications more explicit. Could we keep the library the same and compose the pieces differently?

It looks like the assign vs. charge initial state is actually totally different—meaning we could just use two separate view models e.g.

```javascript
// apps/manage/lib/view-models/assign.js
export default define(Assign.initialState, {
  resetUpdateStatus : reducer(Assign.actions. resetUpdateStatus)
  ...
})

// apps/manage/lib/view-models/charge.js
export default define(Charge.initialState, {
  getCharges : reducer(Charge.actions.getCharges)
})
```

Then if we need to reuse behavior across them for any reason we can compose them from [POJOs](http://blog.dreasgrech.com/2012/02/creating-pojos-in-javascript.html) and pure functions, e.g.

````javascript
// apps/manage/lib/view-models/sidebar.js
export const initialState = {
  sideBarExpanded: true,
  loggedInUser: {
    name: 'Joe'
  }
}

export const toggleSidebar = state => ({
  sideBarExpanded: !state.sideBarExpanded
})

// apps/manage/lib/view-models/assign.js
const initialState = props => ({
  ...Sidebar.initialState,
  ...
})

export default define(initialState, {
  toggleSidebar : reducer(Sidebar.toggleSidebar)
  ...
})
````

Might be a little more verbose but I think the explicitness and not assuming all view models will be composed of subState namespaces is better.

File: apps/manage/lib/view-model/charge.js [hunk 2]
```diff
@@ -0,0 +1,59 @@
...
+  return (await gql.request(query))
+}
+/**
+ * -----------------------------------------------------------------------------
+ * Action:
+ * Actions for components to update global state and/or fetch/mutation data from
+ * external sources
+ * -----------------------------------------------------------------------------
+ */
+export const actions = {
```
craigspaeth: Small nit-pick, but `actions` have a loaded meaning in this part of the stack thanks to Redux. We could probably just export these functions directly without a namespace.

craigspaeth:
Looking good but I think I'd prefer a lighter-weight approach detailed in my first comment. Lmk if I'm overlooking any reason that's not easy to do.

lionsdwarf:
Thanks for the prompt feedback. I'd clearly lost my way and overlooked your simple approach. I guess I got lost in the details 🤷‍♂ . Thank you!

Also noted regarding `actions` namespace.

PR #1343, opened 2019-11-20
Refactor salve notifications

TMcMeans:
Adds salve notifications to the redis bull queue. The fetchJobs array now includes a 'type' property to determine which functions to run to create data rollups based on notification type (appointment or salve).


apps/api/models/salve/domain.js
```diff
@@ -34,14 +34,14 @@ export const toClientResults = (email, results) =>
...
-  JOB_INTERVAL_L = process.env.JOB_INTERVAL_L,
+  timeAgo = moment().subtract(1, 'minute'),
   now = moment()
 ) => {
   const recentResults = salvesResults.map(results =>
     results.filter(({ timestamp }) =>
       momentTz
         .tz(timestamp, 'MMM DD, YYYY hh:mm:ss A', 'America/Los_Angeles')
-        .isAfter(moment(now).subtract(JOB_INTERVAL_L, 'milliseconds'))
+        .isAfter(moment(now).subtract(timeAgo, 'milliseconds'))
```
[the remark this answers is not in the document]
TMcMeans: I think this was mostly done to make the time subtracting more readable?

craigspaeth:
👍 LGTM! @lionsdwarf to your `sendFn` point that had crossed my mind too, but as you noticed this can't work with the queue because (I'm presuming) the data sent [`queue.add`](https://github.com/findOctave/cortex/pull/1343/files#diff-99640e27839994740597c6aeeda396cbR67) has to be POJO JSON data that can be easily serialized to Redis—therefore a function can't be in there because how would you store a function that potentially depends on external state in a database?

That said, we could maybe do something clever with pulling out the string <> sendFn definitions to clean up the code a bit.

apps/api/queue.js
```diff
@@ -10,45 +10,68 @@ export const fetchJobs = [
...
     cron: EVERY_MINUTE,
-    fn: Notifications.rollupNewBookings
+    fn: Notifications.rollupNewBookings,
+    type: 'appointment'
   },
   {
     queue: bull('fetch/apptsUpdatedInLastMin'),
     cron: EVERY_MINUTE,
-    fn: Notifications.rollupRecentlyUpdatedBookings
+    fn: Notifications.rollupRecentlyUpdatedBookings,
```
craigspaeth: I think you need to pull the latest so this remains commented out.

apps/api/models/practitioner/mapper.js
```diff
@@ -68,11 +68,13 @@ export const findForPatient = async patientId => {
     resource: 'practitioners',
     url: Appointment.latestSession(appts, apptTypes).practitioner.links.self
   })
-  return practitioner
+  return toModel(practitioner)
```
craigspaeth: Ideally, we'd add a test to this function if there isn't already one.

craigspaeth:
LGTM. Just need to pull from master, resolve conflicts, and make sure the `rollupRecentlyUpdatedBookings` remains commented out.

apps/api/queue.js
TMcMeans: Either that or deal with the merge conflicts

TMcMeans:
Pulled down master and fixed the merge conflicts. Should be all ready to merge!

PR #1349, opened 2019-11-25
Comment out appts updated in last min in bull queue

TMcMeans:
Commenting out the apptsUpdatedInLastMin jobs so that I can figure out what is being rescheduled vs. updated


craigspaeth:
Thanks for the quick work!

PR #1351, opened 2019-11-25
Comment out function rollup for cancelled appts. for now

craigspaeth:
Because they were rolling up the same as rescheduled appts. causing the confirmations to still send.


apps/api/queue.js
```diff
@@ -18,11 +18,11 @@ export const fetchJobs = [
...
   // },
-  {
-    queue: bull('fetch/apptsCanceledInLastMin'),
-    cron: EVERY_MINUTE,
-    fn: Notifications.rollupRecentlyUpdatedBookings
-  },
+  // {
+  //   queue: bull('fetch/apptsCanceledInLastMin'),
+  //   cron: EVERY_MINUTE,
+  //   fn: Notifications.rollupCanceledBookings
```
craigspaeth: This was using the wrong function, but just in case I'm commenting out so we can QA this.

TMcMeans: That's what I get for trying to PR without reading fast enough

PR #1354, opened 2019-11-25
Turn on canceled appts job in queue

TMcMeans:
Cancelation emails are generating on local, but still need to test in staging.


lib/queue.js
```diff
@@ -13,7 +13,7 @@ const bull = (name, url = REDIS_URL, opts) => {
 
 export const UIQueues = [
   bull('fetch/apptsNewInLastMin'),
-  // bull('fetch/apptsUpdatedInLastMin'),
+  bull('fetch/apptsUpdatedInLastMin'),
```
craigspaeth: We don't want this back in right? Just the line below it

TMcMeans: Oh woops! Editing right now.

TMcMeans: Thank god for code review :)

craigspaeth:
LGTM! Thanks!

PR #1364, opened 2019-11-27
coercenum

lionsdwarf:
I saw that the build failed and think this should resolve the issue. Surprised it didn't occur locally.


craigspaeth:
Thanks for jumping on this quick!

PR #1366, opened 2019-11-27
Or equal to queries to Cliniko in bookings between

craigspaeth:
After testing locally I have a theory that we miss out on reminders because our cron is _so_ reliable that it runs on the second and Cliniko ignores it because we were using greater than and less than in the query params—excluding exact `=` matches.


craigspaeth:
Innocuous—going to test this out on staging's Bull.

PR #1375, opened 2019-12-02
Query appts based on tz guess

lionsdwarf:
This should be good enough for now but it's worth noting that for someone trying to charge appointments across multiple timezones, some late-in-day appts could still appear on the following day. For example, for ET users, late-in-the-PT-day appts get punted to the following day. Let me know if we should adjust the query window back a few hours to accommodate (we can probably get away w this since we don't have early morning appts in any timezone--otherwise to be fully "correct" I think we will need to expand the query window beyond 24 hrs.)


apps/manage/lib/view-model/charge/network.js
```diff
@@ -36,8 +36,8 @@ export const mutationArgs = charge => {
 }
 
 export const queryArgs = selectedDate => {
-  const DT_FORMAT = 'ddd, D MMM YYYY HH:mm:ss z'
-  const date = moment(selectedDate).utc()
+  const DT_FORMAT = 'ddd, D MMM YYYY HH:mm:ss'
```
craigspaeth: `moment.format()` returns a UTC string format by default (and includes some info not present here). Am I missing a reason why we're manually formatting a UTC-like string here?

PR #1376, opened 2019-12-02
utc string for client query

lionsdwarf:


craigspaeth:
LGTM!

apps/manage/test/lib/view-model/charge/network.test.js
```diff
@@ -200,8 +200,8 @@ describe('queryArgs', () => {
   const selectedDate = new Date('Tue Nov 26 2019 10:37:22 GMT-0500')
```
craigspaeth: Make sure your stub data is in UTC string format too.

PR #1378, opened 2019-12-03
Email reliability tweaks

craigspaeth:
Still noticing some emails getting missed and now occasionally duped. This does a couple of things to address that and probably my last attempt at it before I go for a PG-backed solution that persists sent notifications...

- Passes the `job.timestamp` through to run rollups according to the time it was added to the queue vs. when it was processed and potentially delayed
- Rounds the query times to the minute just in case
- Uses >= and < to avoid dupe sending
- Tunes reminders to run at the minute because 🤷‍♂ confirmations are working better and they run on the minute—maybe the smaller Cliniko queries help consistency
- _Scouts Rule_s . by consolidating and cleaning up some code
- Adds a `npm run job clean-queue` job to clean out completed/errored jobs and help debugging


apps/api/models/appointment/domain.js
```diff
@@ -44,3 +44,14 @@ export const latestSession = (appts, apptTypes) =>
     )
     .sortBy(appt => -moment(appt.appointment_start).unix())
     .first()
+
+export const bookingsBetweenQuery = (from, until) => [
+  `starts_at:>=${moment
+    .utc(from)
+    .startOf('minute')
+    .format()}`,
+  `starts_at:<${moment
```
lionsdwarf: 👍

PR #1380, opened 2019-12-03
Sync-clients: Sequentially find users by id instead of fetchAll 

lionsdwarf:


craigspaeth:
Pretty straightforward, thanks!

PR #1383, opened 2019-12-03
30 min cache on Content Modules

lionsdwarf:


apps/api/models/content-module/mapper.js
```diff
@@ -7,7 +7,7 @@ import { airtableToModel } from './domain'
 //  Mapper function called findByName that performs the transaction and calls the airtableToModel function on
 //  the return values.
 
-const cache = { ttl: 1000 * 60 }
+const cache = { ttl: 1000 * 60 * 30 }
```
craigspaeth: Can we add `refreshAhead: true` here too so the cache stays warm?

lionsdwarf: 👍

apps/api/models/content-module/mapper.js
```diff
@@ -7,13 +7,14 @@ import { airtableToModel } from './domain'
...
 
 export const findByName = async name => {
   const records = await at.find({
     table: 'contentModules',
     filter: `name = '${name}'`,
-    cache
+    cache,
+    refreshAhead: true
```
craigspaeth: This is an option [on cache](https://github.com/findOctave/cortex/blob/master/lib/caching-fetch/index.js#L22), so it goes inside the `cache =` part.

lionsdwarf: DOH 🤦‍♂ 
Fixed :)

PR #1384, opened 2019-12-04
Revert "30 min cache on Content Modules"

craigspaeth:
🤦‍♂ so sorry to send you back and forth only to discover that I'm seeing out of memory errors on staging after poking around with this on (not too surprised—refreshing ahead an in-memory cache for all content modules could end up storing a lot of data).

I think if we're going to explore this route we'd need to invest in a more robust caching solution—either HTML-level, GraphQL "data loader" level, or at least storing it in a shared DB with plenty of space like Redis.

For now, mo' cache mo' problems. Twas worth a shot but not worth the investment rn—punt and revert.


lionsdwarf:
Copy that and appreciate the explanation!

lionsdwarf:
Why don't we just kill the 'refeshAhead' but leave the 30 min cache?

PR #1386, opened 2019-12-04
Assign admin role to Cliniko users

lionsdwarf:
When this gets deployed, let's make sure we update the code in the Auth0 Rule and de-activate/delete the to-be-deprecated Authorization Extension.


apps/api/models/user/mapper.js
```diff
@@ -22,8 +23,17 @@ export const findAll = (
 
 export const create = email => auth0.createUser(toAuth0(email))
 
+export const assignRole = (userId, role) =>
+  auth0.assignRolestoUser({ id: userId }, { roles: [role.id] })
+
 export const createUsers = emails => Promise.all(emails.map(create))
 
+export const assignAdminRoles = async users => {
```
craigspaeth: It'll be annoying to mock `auth0` but mind adding an integration test to this function? Then this LGTM 👍

craigspaeth:
Feel free to merge after addressing my test comment. Thanks!

apps/api/models/user/mapper.js
lionsdwarf: Np, will do.

lionsdwarf:
FYI, I added and commented-out the Rule code in Auth0 prod, so flipping the switch will only require uncommenting and deleting the existing role assignment, and then deleting the Authorization integration and Rule.

PR #1388, opened 2019-12-05
Clean queue removes repeatable jobs as well in case of changing cron intervals

craigspaeth:
The issue is that when changing a cron interval the [scheduling cron state stays in Redis](https://github.com/OptimalBits/bull/issues/1005). So when we change it (like we did twice for reminders) it actually adds to that instead of replacing it. This reminds me why I was shy of adding a job queue and avoid database state when possible because _state is the root of all evil_. Code is ephemeral and can be fixed quickly—state bugs creep in hard to find places. It's almost like our queue needs migrations.

In any case, there's this new `npm run job clean-queue` script that I'm going to tack this on to. I think we'll want to keep this adhoc when we discover issues because we don't want to run this stuff on every deploy otherwise it kinda defeats the purpose of the resiliency we get from a queue if we're kludging its state on some automated interval.

All that said, I think the duped email issue should sort itself out eventually going forward but if for some reason it doesn't we can deploy this and run the script.


lionsdwarf:
Oh nice. Yea I remember dealing with the jobs-on-jobs-on-jobs cron issue when we rolled this out. Good find. LGTM!

lionsdwarf:
It's my understanding that, once this is deployed, I should run the `clean-queue` job. If that's not correct, please let me know. @craigspaeth

PR #1390, opened 2019-12-05
Set recipient email from Patient

lionsdwarf:
Salve notifications were failing because they assumed that `email` still lived on `Person`. This remedies the issue. 

Separately but related...I notice the Twilio callout consistently fails due to missing phone numbers (thankfully we already catch the errors). From my testing, it looks like Cliniko API is not returning phone numbers for Patients, but I'd love a sanity check before submitting an issue.


craigspaeth:
LGTM! Thanks!

PR #1392, opened 2019-12-06
Billing page: disable table sort, order charges by location

lionsdwarf:
The sort-by functionality would reorder the rows but the Amounts remained static. We may need to abandon our hacky use of the react-data-table-component, but in the meantime, this should get the page to a usable state with the records ordered by location for the Ops team.


apps/manage/lib/view-model/charge/domain.js
```diff
@@ -1,5 +1,7 @@
 import _ from 'lodash'
 
+export const sortByLocation = charges => _.orderBy(charges, charge => charge.appointment.location.name)
```
craigspaeth: Ideally this has a test

craigspaeth:
LGTM but I'm not seeing the part where we disabled the sorting built-in?

I can totally see how an uncontrolled input wouldn't re-render correctly thought. Have you tried using a `key` on those inputs? Probably wouldn't help b/c we need to value to change with state—hence controlled input.

apps/manage/lib/view-model/charge/network.js
```diff
@@ -95,7 +95,7 @@ export const getChargesForDate = async (state, selectedDate) => {
     return {
       ...state,
       isFetching: false,
-      charges: charges.map(initAmount)
+      charges: sortByLocation(charges.map(initAmount))
```
craigspaeth: Oh I see... you're sorting in the mutation. Interesting. I guess i'm not visualizing the bug, but I trust this is a good solution to it.

apps/manage/components/charge/index.js
```diff
@@ -81,7 +81,7 @@ const tableConfig = [
     selector: 'isSelected',
     cell: row => <RowSelect {...row} key={row.id} />
   }
-].map(col => ({ ...col, sortable: true }))
```
lionsdwarf: @craigspaeth This is where we disable the sorting.

apps/manage/lib/view-model/charge/domain.js
lionsdwarf: I can add a test but isn't it just testing a lodash function ?

apps/manage/lib/view-model/charge/network.js
lionsdwarf: We sort by location now that the columns are no longer sortable.

lionsdwarf:
Unfortunately using `key`s did not solve the issue.

PR #1395, opened 2019-12-06
fix person update to preserve phone num

lionsdwarf:


lionsdwarf:
Merging since we reviewed IRL

PR #1400, opened 2019-12-10
Toggle job via env var

lionsdwarf:


apps/api/jobs/charge-sessions/index.js
```diff
@@ -18,29 +18,32 @@ const twoHours = 1000 * 60 * 60 * 2
 const twoWeeksAgo = moment().subtract(2, 'weeks')
 const sixMonthsAgo = moment().subtract(6, 'months')
 
-export default () =>
-  Status.preventConcurrency('charge-sessions', async () => {
-    const recentAppts = await getApptsWithBillables(twoWeeksAgo, moment())
-    // Warn of charges and missing credit cards to surface anomolies
+export default () => {
+  if (process.env.IS_CHARGING_ACTIVE === 'true') {
```
craigspaeth: Pro tip: This is a good place for an early return. A general rule of thumb I like to follow is—nesting hurt readability because you have to read it as if (...read a large block of code and keep it in your head...) else (...what was that thing in my head) vs. if (blah) return ... okay now I can forget that and read the rest of it with free brainspace.

````javascript
if (process.env.IS_CHARGING_ACTIVE !== 'true') return
````

craigspaeth:
LGTM! Thanks.

